PR #417, opened 2019-03-01
Add a tag to latest images to trigger Flux deployments

slyg:
This PR is related to the implementation of flux and allows it to trigger a deployment in the aat AKS cluster when any new master image is pushed in the hmcts registry.

The known drawback is that it'll create loads of tags but I assume we'll create or re-use the existing images cleanup scripts.

Typically this modification means that any image built with the following name:

```
<image-name>:latest
```

will also be tagged as

```
<image-name>:latest-dfb02 # corresponding to the {{.Run.ID}} of the acr build task
```

There is no other functional change, and should not have any impact on the existing pipelines. It just generates an additional tag triggering a deployment in the aat aks cluster if Flux is in use and expecting this pattern.

### DIFF
--- a/src/uk/gov/hmcts/contino/azure/Acr.groovy
+++ b/src/uk/gov/hmcts/contino/azure/Acr.groovy
@@ -55,7 +55,7 @@ class Acr extends Az {
    * Notice that when an image is tagged as 'latest', another tag
    * with a suffix appended 'latest' is also created to trigger
    * possible downstream deployments:
-   * e.g. <image-name>:latest will also be tagged <image-name>:latest~dfb02
+   * e.g. <image-name>:latest will also be tagged <image-name>:latest_dfb02
    *
    * @param dockerImage
    *   the docker image to build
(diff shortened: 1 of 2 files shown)


src/uk/gov/hmcts/contino/azure/Acr.groovy
```diff
@@ -52,14 +52,20 @@ class Acr extends Az {
...
+   *
    * @param dockerImage
    *   the docker image to build
    *
    * @return
    *   stdout of the step
    */
   def build(DockerImage dockerImage) {
-    this.az "acr build --no-format -r ${registryName} -t ${dockerImage.getShortName()} -g ${resourceGroup} ."
+    def additionalTag = (dockerImage.getTag() != "latest") ? "" : "-t ${dockerImage.getShortName()}_{{.Run.ID}} "
```
timwebster9: nitpicking here, but maybe `-` instead of `_` is more conventional?

slyg: OK sure np

timja:
What does it get tagged with?
i.e. can you add an example, not sure what shortName returns

Also I don't think the jenkins-library is currently running the tests for docker because of the issue with plum not being able to pull from the hmcts registry, so this should be tested on a separate PR or the docker issue fixed so it can be re-enabled in the pipeline

slyg:
@timja I've updated the PR description. an example is also in the code comments.

timja:
have you got a link to somewhere you've tested it?

slyg:
Here is a [failed build](https://sandbox-build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_Sandbox_Pipeline_Test%2Fcnp-plum-recipes-service/detail/master/36/pipeline) (because of invalid character use '~')

And the [fixed one](https://sandbox-build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_Sandbox_Pipeline_Test%2Fcnp-plum-recipes-service/detail/master/38/pipeline/47) obviously.

So I assume that the acr build is being tested in this PR's checks.

timja:
Ah its being tests on backend but not frontend 👍 
thanks

PR #419, opened 2019-03-05
Refactor PipelineCallbacks

njrich28:
This has become a huge class with lots of different concerns.

Split class into 4, each with separate concern
 * AppPipelineConfig - container for app pipeline config options
 * PipelineCallbacksConfig - container for pipeline callback blocks
 * AppPipelineDsl - DLS methods available to users of `withPipeline` and registering callbacks
 * PipelineCallbacksRunner - helper wrapper to safely execute callbacks


src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -11,7 +11,7 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
 
   @Deprecated
   void loadVaultSecrets(List<Map<String, Object>> vaultSecrets) {
-    this.vaultSecrets = ['unknown': vaultSecrets]
+    config.vaultSecrets = ['unknown': vaultSecrets]
```
timja: we should look too remove this at some point i guess, don't want to maintain it forever

PR #423, opened 2019-03-08
Add docs for withInfraPipeline

njrich28:


README.md
```diff
@@ -54,6 +54,15 @@ withPipeline(type, product, component) {
...
 ```
 
+The opinionated pipeline uses the following branch mapping to deploy applications to different environments.
+
+Branch | Environment
+--- | ---
+`master` | `aat` then `prod`
+`demo` | `demo`
+`perftest` | `perftest`
+PR branch| `preview` (ASE or AKS depending on your config)
```
timja: does this deploy or run plan only?
preview doesn't work very well for shared infra because it gets deleted constantly, for teams like bulk-scanning it's a pain in the butt because they need devops intervention 

what was being done instead in their pipeline is running plan against AAT...

README.md
```diff
@@ -199,6 +208,64 @@ tests for that API. For the pipeline to run those tests, do the following:
...
+For infrastructure-only repositories e.g. "shared infrastructure" the library provides an opinionated infrastructure pipeline which will build Terraform files in the root of the repository.
+
+It uses a similar branch --> environment strategy as the app pipeline but with some differences for PRs
+
+Branch | Environment
+--- | ---
+`master` | `aat` then `prod`
+`demo` | `demo`
+`perftest` | `perftest`
+PR branch| `aat` (plan only)
```
timja: 👍

README.md
```diff
@@ -199,6 +208,64 @@ tests for that API. For the pipeline to run those tests, do the following:
...
+withInfraPipeline(product) {
+}
+```
+
+#### Slack notifications on failure / fixed
+
+To enable slack notifications when the build fails or is fixed add the following:
+```groovy
+withInfraPipeline(product) {
+  enableSlackNotifications('#my-team-builds')
```
timja: is it still needed to keep slack notifications in a separate section?
I would always want them enabled, maybe we could remove this block and add it to the main docs??

njrich28: Yeah was thinking similar - i'll extract out

PR #431, opened 2019-03-20
Git image tag

luigibk:
https://tools.hmcts.net/jira/browse/RPE-1010

Tag docker images using git commit hash

### DIFF
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -20,8 +20,11 @@ def call(params) {
     pcr.callAround('checkout') {
       deleteDir()
       def scmVars = checkout scm
-      params.commit = scmVars.GIT_COMMIT
-      params.branch = scmVars.GIT_BRANCH
+      if (scmVars) {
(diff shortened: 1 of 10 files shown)


vars/sectionBuildAndTest.groovy
```diff
@@ -19,7 +19,12 @@ def call(params) {
   stage('Checkout') {
     pcr.callAround('checkout') {
       deleteDir()
-      checkout scm
+      def scmVars = checkout scm
+      if (scmVars) {
+        env.GIT_COMMIT = scmVars.GIT_COMMIT
+        env.GIT_BRANCH = scmVars.GIT_BRANCH
```
timja: This "should" already be exported as `CHANGE_BRANCH` assuming that is available ideally we should use the standard variables and not overwrite.

luigibk: CHANGE_BRANCH (git-image-tag) is a different variable from GIT_BRANCH (PR-431)

To be honest I only need GIT_COMMIT. I thought it would be useful to export the entire output of the checkout operation, but it looks like that might be confusing. I have removed all the rest.

PR #432, opened 2019-03-20
updated track your appeal name

chrisdavecm:
Notes: tya had its name changed, updating to current value to fix helm chart deployment

### DIFF
--- a/src/uk/gov/hmcts/contino/TeamNames.groovy
+++ b/src/uk/gov/hmcts/contino/TeamNames.groovy
@@ -37,7 +37,7 @@ class TeamNames {
                       'plum-shared-infrastructure' : 'CNP',
                       'sscs':'SSCS',
                       'sscs-cor':'SSCS',
-                      'sscs-tya':'SSCS',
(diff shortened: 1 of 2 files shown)


src/uk/gov/hmcts/contino/TeamNames.groovy
```diff
@@ -37,7 +37,7 @@ class TeamNames {
                       'plum-shared-infrastructure' : 'CNP',
                       'sscs':'SSCS',
                       'sscs-cor':'SSCS',
-                      'sscs-tya':'SSCS',
+                      'sscs-track-your-appeal-frontend':'SSCS',
```
timja: that's not a product name, that's an application name.
Changing a product name will cause the dns name to change for an application, it's not something we normally do, why is it being changed?

normally a product would be something like:
`sscs`

if a complete separate team is running it then it might be:
`sscs-tya`
`tya`

`track-your-appeal`

`sscs-track-your-appeal`

but including frontend makes it a `component`

PR #441, opened 2019-04-05
Build image after tests/checks

luigibk:

https://tools.hmcts.net/jira/browse/RPE-1037

Build image cannot be parallel anymore, otherwise there might be cases of unsuccessful stages (e.g. tests) that are skipped. 

Notes:
*
*
*

### DIFF
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -47,7 +47,7 @@ def call(params) {
     }
   }
 
-  stage("Tests/Checks/Container build") {
+  stage("Tests/Checks") {
     when (tagMissing) {
       parallel(
         "Unit tests and Sonar scan": {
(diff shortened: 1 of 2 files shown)


vars/sectionBuildAndTest.groovy
```diff
@@ -47,7 +47,7 @@ def call(params) {
     }
   }
 
-  stage("Tests/Checks/Container build") {
+  stage("Tests/Checks") {
```
doncem: is it possible to somehow extend the tests/checks step to run multiple different tests in parallel?

slyg:
Oh but this feature allowed time gains in the build, sure there is no other way to maintain parallelisation ?

vars/sectionBuildAndTest.groovy
luigibk: Maybe. What tests do you have in mind?

doncem: Well typically we have unit tests and integration tests. They are independent and can run in parallel. No doubt there can be more at this stage

slyg: Integration tests might need a deployment. However loosing the parallelisation make some project build longer, it was the whole point of this parallelisation. e.g. (I count the shortest)

div-decree-nisi-frontend 1:35 min
probate-frontend 3:30 min
probate-caveat-frontend 1:20 min
div-case-orchestration-service 3:00 min
div-petitioner-frontend 2:20 min

doncem: perhaps up to a project. normally integration tests don't need deployment. functional tests do. i was more on topic whether individual project incorporating this jenkins library can customly extend this stage/step by including additional parallel functions

slyg: Oh sure ok. But as is, these steps can't be extended easily... not sure this even possible.

PR #445, opened 2019-04-08
Updated cve validity to 48 hours

nitinprabhu:
Notes:
* Currently CVE validity is 24 hours and this PR updates it to 48 hours.
* This PR is a temporary resolution to download the CVE's only after 48 hours as currently there is a problem where downloading of CVE's fail intermittently and the build are completely blocked.
* Platform team is working on a more strategic which once done then this change can be reverted.


timja:
Same as: https://github.com/hmcts/cnp-jenkins-library/pull/444 ?

PR #450, opened 2019-04-10
No skip build pr

luigibk:

PR builds can now be skipped as the frontend build checks and install dependencies before running tests.

See: https://tools.hmcts.net/jira/browse/RPE-1037

Notes:
*
*
*

### DIFF
--- a/src/uk/gov/hmcts/contino/YarnBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/YarnBuilder.groovy
@@ -104,7 +104,7 @@ EOF
     steps.sh("yarn ${task}")
   }
 
-  def yarnQuiet(task){
+  def yarnQuiet(task) {
     return steps.sh(script: "yarn ${task} &> /dev/null", returnStatus: true)
   }
 
--- a/src/uk/gov/hmcts/contino/YarnBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/YarnBuilder.groovy
@@ -9,20 +9,19 @@ class YarnBuilder extends AbstractBuilder {
   }
 
   def build() {
-    runYarn("--mutex network install --frozen-lockfile")
-    runYarn("lint")
+    yarn("lint")
 
     addVersionInfo()
   }
 
   def test() {
-    runYarn("test")
-    runYarn("test:coverage")
-    runYarn("test:a11y")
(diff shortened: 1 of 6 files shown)


src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -101,4 +103,17 @@ EOF
   def yarn(task){
     steps.sh("yarn ${task}")
   }
+
+  def yarnQuiet(task){
+    return steps.sh(script: "yarn ${task} &> /dev/null", returnStatus: true)
+  }
+
+  def yarnWithCheck(task) {
```
timja: This doesn't read well to me,

can't we keep it the same:
```suggestion
  def yarn(task) {
```

and then make a private function, something like:
```
private runYarn(task) {
```

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -101,4 +103,17 @@ EOF
   def yarn(task){
     steps.sh("yarn ${task}")
   }
+
+  def yarnQuiet(task){
```
timja:
```suggestion
  def yarnQuiet(task) {
```

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -2,25 +2,27 @@ package uk.gov.hmcts.contino;
...
+  private static final String INSTALL_CHECK_FILE = '.yarn_dependencies_installed'
+
   YarnBuilder(steps) {
     super(steps)
   }
 
   def build() {
-    yarn("--mutex network install --frozen-lockfile")
-    yarn("lint")
+    runYarn("--mutex network install --frozen-lockfile")
```
timja: shouldn't all these methods call `yarn`?

you could then get rid of this line as it will get invoked by the lint call

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -2,25 +2,27 @@ package uk.gov.hmcts.contino;
...
+    runYarn("lint")
```
timja:
```suggestion
    yarn("lint")
```

PR #455, opened 2019-04-15
Refactoring AKS config to read from env Vars

adusumillipraveen:
Notes:
* Refactoring AKS cluster name and resource group to read from env Vars 
* Added default names dynamically for consistency.

### DIFF
--- a/vars/aksDeploy.groovy
+++ b/vars/aksDeploy.groovy
@@ -7,7 +7,7 @@ import uk.gov.hmcts.contino.Consul
 def call(DockerImage dockerImage, Map params) {
 
   def subscription = params.subscription
-  def environment = params.parentEnvionment
+  def environment = params.parentEnvironment
 
   def kubeResourcesDir
   def kubeResourcesDirDefault = "src/kubernetes"
--- a/vars/aksDeploy.groovy
+++ b/vars/aksDeploy.groovy
@@ -7,7 +7,7 @@ import uk.gov.hmcts.contino.Consul
 def call(DockerImage dockerImage, Map params) {
 
   def subscription = params.subscription
-  def environment = params.environment
+  def environment = params.parentEnvionment
 
   def kubeResourcesDir
   def kubeResourcesDirDefault = "src/kubernetes"
--- a/vars/aksDeploy.groovy
+++ b/vars/aksDeploy.groovy
@@ -3,11 +3,15 @@ import uk.gov.hmcts.contino.HealthChecker
 import uk.gov.hmcts.contino.Kubectl
 import uk.gov.hmcts.contino.GithubAPI
 import uk.gov.hmcts.contino.Consul
+import uk.gov.hmcts.contino.Environment
 
 def call(DockerImage dockerImage, Map params) {
 
   def subscription = params.subscription
-  def environment = params.parentEnvironment
+  def environment = params.environment
+  onPR {
+    environment = new Environment(env).nonProdName;
+  }
 
   def kubeResourcesDir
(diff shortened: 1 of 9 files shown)


vars/sectionCI.groovy
```diff
@@ -27,19 +27,21 @@ def call(params) {
   def product = params.product
   def component = params.component
   def aksUrl
+  def environment = params.environment;
+  def parentEnvionment = params.parentEnvironment; //Used for secrets which are maintained only in parent environments.
```
timja:
```suggestion
  def parentEnvironment = params.parentEnvironment; // Used for secrets which are maintained only in parent environments.
```

vars/withAksClient.groovy
```diff
@@ -18,10 +16,14 @@ def withRegistrySecrets(Closure block) {
...
 
 
-def call(String subscription, Closure block) {
+def call(String subscription, String environment, Closure block) {
   withDocker('hmcts/cnp-aks-client:az-2.0.50-kubectl-1.12.2-helm-2.11', null) {
     withSubscription(subscription) {
       withRegistrySecrets {
+        def envName = environment.toUpperCase()
+        echo "envname : ${envName} "
+        env.AKS_CLUSTER_NAME = env."${envName}_AKS_CLUSTER_NAME" ?: "cnp-${environment}-cluster"
```
timja: do we need the fallbacks here?
can't we just rely on the env variables always being set?

adusumillipraveen: Ideally if we had good naming convention, we wouldn't need have to set them in environment. I would rather avoid the need of setting them  by following a convention.

vars/aksDeploy.groovy
```diff
@@ -7,7 +7,7 @@ import uk.gov.hmcts.contino.Consul
 def call(DockerImage dockerImage, Map params) {
 
   def subscription = params.subscription
-  def environment = params.environment
+  def environment = params.parentEnvionment
```
luigibk: params.parentEnvironment

vars/sectionCI.groovy
```diff
@@ -27,19 +27,21 @@ def call(params) {
...
+    withAksClient(subscription,environment) {
       def acr = new Acr(this, subscription, env.REGISTRY_NAME, env.REGISTRY_RESOURCE_GROUP)
       def dockerImage = new DockerImage(product, component, acr, new ProjectBranch(env.BRANCH_NAME).imageTag(), env.GIT_COMMIT)
 
       onPR {
         acr.retagForStage(DockerImage.DeploymentStage.PR, dockerImage)
 
         if (config.deployToAKS) {
-          withTeamSecrets(config, params.environment) {
+          withTeamSecrets(config, parentEnvironment) {
```
njrich28: `withTeamSecrets` has built in support for overriding vaults between environments https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/withTeamSecrets.groovy#L46.

The `config` object contains a map of environments (configurable by teams) to override the environment to use for vaults. The map defaults to empty `[:]` (i.e. no overrides) but we could default it to `['preview':'aat']`

vars/sectionCI.groovy
```diff
@@ -27,19 +27,20 @@ def call(params) {
   def product = params.product
   def component = params.component
   def aksUrl
+  def environment = params.environment;
 
   Builder builder = pipelineType.builder
 
   if (config.dockerBuild) {
-    withAksClient(subscription) {
+    withAksClient(subscription,environment) {
```
timja:
```suggestion
    withAksClient(subscription, environment) {
```

PR #471, opened 2019-04-29
Making consul rg name dynamic based on environmentname

adusumillipraveen:
Notes:
* Making consul rg name dynamic based on environmentname

### DIFF
--- a/src/uk/gov/hmcts/contino/Consul.groovy
+++ b/src/uk/gov/hmcts/contino/Consul.groovy
@@ -12,10 +12,10 @@ class Consul {
   def az
   private consulApiAddr
 
-  Consul(steps) {
+  Consul(steps, environment) {
     this.steps = steps
     this.subscription = this.steps.env.SUBSCRIPTION_NAME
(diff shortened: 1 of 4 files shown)


src/uk/gov/hmcts/contino/Consul.groovy
```diff
@@ -8,12 +8,14 @@ class Consul {
 
   def steps
   def subscription
+  def environment
   def az
   private consulApiAddr
 
-  Consul(steps) {
+  Consul(steps, environment) {
```
timja: can you check github for any uses of this class outside the library?

adusumillipraveen: Thanks, There is one place where they are using. Should i make it backward compatible with default env as preview? 
https://github.com/hmcts/probate-intestacy-frontend/blob/56d7a9350059232e9213b358084f779ad91f17b9/Jenkinsfile_CNP#L76

timja: yup, nice work

PR #473, opened 2019-04-30
Fixing skip builds on master branches

adusumillipraveen:
Notes:
* Fixing skip builds on master branches as currently some of the builds are incorrectly skipping build tests
* Adding an extra stage AAT doesn't seem avoidable for now, will need to refactor this in future.


src/uk/gov/hmcts/contino/azure/Acr.groovy
```diff
@@ -135,7 +135,7 @@ class Acr extends Az {
 
   private def hasRepoTag(String tag, String repository) {
     // latest is not really a tag for our purposes, it just marks the most recent tag
```
timja: comment wrong?

PR #492, opened 2019-05-21
Fix skip build issue on master

adusumillipraveen:
Notes:
*Change condition as images are now tagged as staging- commit and not staging .


timja:
We really need a better way of testing this, its been broken soo many times :(

PR #493, opened 2019-06-06
Differentiating withAksClient from ACR client

adusumillipraveen:
Resolves [AB#177](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/177)

Notes:
* Differentiating withAksClient from ACR client as acr and aks will be in different subscriptions from now on.
* This is a step towards moving towards new subscriptions and doing in small PRs for simplicity.

### DIFF
--- a/vars/withAcrClient.groovy
+++ b/vars/withAcrClient.groovy
@@ -1,0 +1,30 @@
+import uk.gov.hmcts.contino.Environment
+
+def withRegistrySecrets(Closure block) {
+  def registrySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'registry-name', version: '', envVariable: 'REGISTRY_NAME'],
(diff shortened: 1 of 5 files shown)


vars/withAcrClient.groovy
```diff
@@ -0,0 +1,30 @@
+import uk.gov.hmcts.contino.Environment
+
+def withRegistrySecrets(Closure block) {
+  def registrySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'registry-name', version: '', envVariable: 'REGISTRY_NAME'],
```
timja: I would consider moving these out of keyvault, it doesn't really add any value.
if anything they should be env vars on jenkins

adusumillipraveen: As discussed, will leave this for AB#95 to handle.

PR #500, opened 2019-07-03
Add retry on helm commands

timja:
Should help if tiller is being flakey...


timja:
Azure is failing, the functional change has been test properly

PR #505, opened 2019-07-10
Register dns selectively for ASE or AKS (or both)

luigibk:

[AB#309]


vars/helmPublish.groovy
```diff
@@ -40,6 +42,13 @@ def call(Map params) {
       }
 
       helm.publishIfNotExists(values)
+
+      if (!config.legacyDeployment) {
```
timja: is this check the wrong way round?

timja: actually I think it's ok

vars/registerDns.groovy
```diff
@@ -0,0 +1,53 @@
...
+      } else {
+        echo "Skipping dns registration for AKS as this environment is not configured with it: ${aksSubscriptionName}"
+      }
+    }
+    // AAT + PROD DNS registration
+    def aksEnv = params.aksSubscription != null && params.aksSubscription.envName
+
+    if (config.legacyDeployment) {
+      withIlbIp(params.environment) {
+        consul.registerDns("${params.product}-${params.component}-${params.environment}", env.TF_VAR_ilbIp)
```
njrich28: Can this be moved to line 14 (after the `-staging` registration)? Save the cal to reasonably expensive `withIlbIp`

PR #509, opened 2019-07-16
Moving team names to cnp-jenkins-config 

adusumillipraveen:
Notes:
* Fixes  https://github.com/hmcts/cnp-jenkins-library/issues/413
*
*

### DIFF
--- a/src/uk/gov/hmcts/contino/TeamNames.groovy
+++ b/src/uk/gov/hmcts/contino/TeamNames.groovy
@@ -22,7 +22,7 @@ class TeamNames {
         url: "https://raw.githubusercontent.com/hmcts/cnp-jenkins-config/master/team-config.yml",
         validResponseCodes: '200'
       )
-      teamNamesMap = steps.readYaml (text: response)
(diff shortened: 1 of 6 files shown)


src/uk/gov/hmcts/contino/TeamNames.groovy
```diff
@@ -2,77 +2,59 @@ package uk.gov.hmcts.contino
...
+    this.steps = steps
+  }
+
+  def getTeamNamesMap() {
+    if (teamNamesMap ==null ){
+      def response = steps.httpRequest(
+        consoleLogResponseBody: true,
+        authentication: "${GITHUB_CREDENTIAL}",
+        timeout: 10,
+        url: "https://raw.githubusercontent.com/hmcts/cnp-jenkins-config/master/team-config.yml",
```
timja: this PR looks great, only question would be is that the most appropriate place?
i.e. should we use a generic config repo, something like:
azure-team-config ? @njrich28

adusumillipraveen: Decided to continue in there for now.

PR #510, opened 2019-07-17
Delete dir + few more

timja:
Fixes AB#410

I've added a few changes here, separated out into different commits

* Remove timestamps step, as every build is now logging a warning saying it's unneeded now
* Use unstable builds in nightly rather than rolling our own hack around it (see https://jenkins.io/blog/2019/07/05/jenkins-pipeline-stage-result-visualization-improvements/)
* Delete old pipeline that's not in use
* Cleanup workspaces at the end of builds


Jenkinsfile_pipeline_test
```diff
@@ -50,6 +50,8 @@ node {
   } catch (err) {
     notifyBuildFailure channel: channel
     throw err
+  } finally {
+    deleteDir()
```
adusumillipraveen: Should we remove it from checkout step ? Or may be just let it be there as there is no harm?

adusumillipraveen: I think always good to have it during checkout. Forget this.

timja:
image

image

PR #512, opened 2019-07-18
Fix FilePath error in nightly pipeline

doncem:
Similarly like #511, fix is rolled in nightly pipeline build too


timja:
Overriding check as build failure is unrelated, there's no coverage for this pipeline in tests...

PR #519, opened 2019-07-31
Add approval check for infrabuild and deploy to environments

njrich28:


src/uk/gov/hmcts/pipeline/EnvironmentApprovals.groovy
```diff
@@ -0,0 +1,37 @@
+package uk.gov.hmcts.pipeline
+
+class EnvironmentApprovals {
+  def steps
+  static final String GITHUB_CREDENTIAL = 'jenkins-github-hmcts-api-token'
+  static def approvalsMap
```
timja: shouldn't need to include the type in a variable name?
if you're tempted to do that and IDE can't do it for you maybe put the actual type in groovy?
I assume it's

```suggestion
  private static Map<String, Object> approvalsMap
```

vars/sectionDeployToEnvironment.groovy
```diff
@@ -21,66 +21,68 @@ def call(params) {
   Deployer deployer = pipelineType.deployer
   def tfOutput
 
-  lock(resource: "${product}-${component}-${environment}-deploy", inversePrecedence: true) {
-    folderExists('infrastructure') {
-      stage("Build Infrastructure - ${environment}") {
-        onPreview {
-          deploymentNumber = githubCreateDeployment()
-        }
+  approvedEnvironmentReposiitory(environment) {
```
timja:
```suggestion
  approvedEnvironmentRepository(environment) {
```

src/uk/gov/hmcts/pipeline/EnvironmentApprovals.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+  boolean isApproved(String environment, String repository) {
+    def environmentApprovalsMap = getEnvironmentApprovalsMap()
+    if (!environmentApprovalsMap.containsKey(environment)) {
+      // if environment is not listed in config then assume no whitelist required
+      return true
+    }
+
+    def approvals = environmentApprovalsMap.get(environment)
+    return approvals.findResult(false) { it.repo.equalsIgnoreCase(repository) ? true : null }
```
timja: If we do this it's going to be a complete breaking change.
few options I guess,
1. just break it
2. Implement SlackNagger (worthwhile imo)
3. do a best efforts pre-fill it with what's currently in prod and break it

src/uk/gov/hmcts/pipeline/EnvironmentApprovals.groovy
```diff
@@ -0,0 +1,37 @@
+package uk.gov.hmcts.pipeline
+
+class EnvironmentApprovals {
```
timja: Probably worth a test? Should be able to lift most of it from the TeamNames test?

src/uk/gov/hmcts/pipeline/EnvironmentApprovals.groovy
njrich28: cut'n'paste from TeamNames 😄 - probably better called environmentApprovals anyway

src/uk/gov/hmcts/pipeline/EnvironmentApprovals.groovy
njrich28: So, start with 2 (I'll actually change to make it non-breaking) and warn/notify people
Then 3
Then 1

vars/approvedEnvironmentRepository.groovy
```diff
@@ -10,9 +10,23 @@ import uk.gov.hmcts.pipeline.EnvironmentApprovals
  * }
  */
 def call(String environment, Closure block) {
-  if (new EnvironmentApprovals(this).isApproved(environment, env.GIT_URL)) {
-    return block.call()
-  } else {
-    echo "Repo ${env.GIT_URL} is not approved for environment ${environment}"
+  if (!new EnvironmentApprovals(this).isApproved(environment, env.GIT_URL)) {
+    echo '''
+================================================================================
```
timja: gonna do the slack nagger as well?

PR #533, opened 2019-08-21
Feature: separate vault value for db migration

doncem:
In order to migrate from deprecated features, particularly [this](https://github.com/hmcts/cnp-jenkins-library#secrets-for-functional--smoke-testing) guideline on _howto_, some projects might use multiple vaults to load secrets - as suggested in the link previously. This feature requires `vaultName` to be removed from terraform output file. Which then causes null pointer in case project has db migration step in the pipeline.

Introducing separate config just for that and securing from null value

### DIFF
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -30,8 +30,10 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
   }
 
   void enableDbMigration(String dbMigrationVaultName = "") {
-   if (dbMigrationVaultName == "") {
-     WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration(<vault-name>)", new Date().parse("dd.MM.yyyy", "05.09.2019"))
+    if (dbMigrationVaultName == "") {
+      WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration(<vault-name>)", new Date().parse("dd.MM.yyyy", "05.09.2019"))
+    }
+
     config.migrateDb = true
     config.dbMigrationVaultName = dbMigrationVaultName
   }
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -30,6 +30,7 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
   }
 
   void enableDbMigration(String dbMigrationVaultName = "") {
+    WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration("<vault-name>")", new Date().parse("dd.MM.yyyy", "05.09.2019"))  
     config.migrateDb = true
     config.dbMigrationVaultName = dbMigrationVaultName
   }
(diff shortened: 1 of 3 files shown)


vars/sectionDeployToEnvironment.groovy
```diff
@@ -52,7 +52,10 @@ def call(params) {
             if (config.migrateDb) {
               stage("DB Migration - ${environment}") {
                 pcr.callAround("dbmigrate:${environment}") {
-                  builder.dbMigrate(tfOutput.vaultName.value, tfOutput.microserviceName.value)
+                  builder.dbMigrate(
+                    tfOutput.vaultName ? tfOutput.vaultName.value : config.dbMigrationVaultName,
+                    tfOutput.microserviceName.value
```
timja: hmm we need to do something about this as well so we can stop relying on the tfoutput,
shouldn't this just be the component parameter?

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -29,8 +29,9 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     WarningCollector.addPipelineWarning("deprecated_set_vault_name", "setVaultName() is deprecated, see https://github.com/hmcts/cnp-jenkins-library#secrets-for-functional--smoke-testing ", new Date().parse("dd.MM.yyyy", "27.08.2019"))
   }
 
-  void enableDbMigration() {
+  void enableDbMigration(String dbMigrationVaultName = "") {
```
adusumillipraveen: Should we make it a non-breaking change, just add it to pipelinewarnings ? something like : https://github.com/hmcts/cnp-jenkins-library/blob/a50fd5216c1b42064a9557e59b479509bff91d27/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy#L29

timja: isn't this non breaking already? a default has been added

adusumillipraveen: Sorry, i meant "auto-bombing" and "deprecation warning" or do we plan to support both ?

timja: yeah makes sense to add it 👍

doncem:
Feedback: DB migration step has passed, [link here](https://sandbox-build.platform.hmcts.net/job/HMCTS_Sandbox_Platform/job/send-letter-service/job/tmp%252Ffix/3)

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
timja:
```suggestion
  void enableDbMigration(String dbMigrationVaultName = "") {
    WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration("<vault-name>")", new Date().parse("dd.MM.yyyy", "05.09.2019"))  
```

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -29,8 +29,10 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
...
+    WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration(<vault-name>)", new Date().parse("dd.MM.yyyy", "05.09.2019"))
```
timja: this needs an if statement around it
 ```suggestion
    if (dbMigrationVaultName == "") {
      WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration(<vault-name>)", new Date().parse("dd.MM.yyyy", "05.09.2019"))
```

doncem: aye, thanks

PR #535, opened 2019-08-23
Restore compatibility for microservice name

timja:
for people that are overriding the microservice name with something else,

add deprecation warning as well...


timja:
image

It outputs twice because db migration runs twice, but w/e i guess?

PR #537, opened 2019-09-05
Fix tests for deprecated functionalities

luigibk:

Notes:
*
*
*


luigibk:
> Or do we get rid of the code altogether ?

I generally prefer to do this after a while just to be cautious. What about adding a clean up ticket to the board (I think there's other deprecated code to remove as well)?

adusumillipraveen:
> > Or do we get rid of the code altogether ?
> 
> I generally prefer to do this after a while just to be cautious. What about adding a clean up ticket to the board (I think there's other deprecated code to remove as well)?

sure, there are tickets for this anyway 👍

PR #538, opened 2019-09-05
SIDM-3120 Allow for unique names for pull request migrations

nikola-naydenov-hmcts:
Currently in idam our pull requests have unique databases but do not have a unique vault, they share the same vault for their secrets.

We currently cannot build our pull requests as all of our secrets follow **product-component-POSTGRES-ETC** format

This should allow us to maintain this functionality with the removal of microserviceName output

### DIFF
--- a/vars/sectionDeployToEnvironment.groovy
+++ b/vars/sectionDeployToEnvironment.groovy
@@ -60,7 +60,7 @@ def call(params) {
                   
                   builder.dbMigrate(
                     tfOutput.vaultName ? tfOutput.vaultName.value : "${config.dbMigrationVaultName}-${environment}",
-                    tfOutput.microserviceName? tfOutput.microserviceName.value : component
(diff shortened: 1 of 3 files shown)


vars/sectionDeployToEnvironment.groovy
```diff
@@ -60,7 +60,7 @@ def call(params) {
                   
                   builder.dbMigrate(
                     tfOutput.vaultName ? tfOutput.vaultName.value : "${config.dbMigrationVaultName}-${environment}",
-                    tfOutput.microserviceName? tfOutput.microserviceName.value : component
+                    config.uniqueDbMigrateSecretNames? "${product}-${component}" : component
```
timja: would rather auto prefix with PR number if on preview, but let's have a chat about this change?

nikola-naydenov-hmcts: sure

PR #552, opened 2019-10-22
Upgrade helm and kubectl

timja:


timja:
Will hold on this for now

PR #553, opened 2019-10-25
Quick fix for sandbox testing

apereverzin:
Resolves # 
https://tools.hmcts.net/jira/browse/BPS-851

Notes:
* Need to fix bug in deprecated code to resolve the task above
*
*

### DIFF
--- a/vars/sharedInfrastructurePipeline.groovy
+++ b/vars/sharedInfrastructurePipeline.groovy
@@ -29,8 +29,8 @@ ______                              _           _ _
 ================================================================================
 '''
 
-  try {
-    node {
+  node {
+    try {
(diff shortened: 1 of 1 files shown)


vars/sharedInfrastructurePipeline.groovy
```diff
@@ -29,8 +29,8 @@ ______                              _           _ _
 ================================================================================
 '''
 
-  try {
-    node {
+  node {
```
timja: should really just not use this pipeline instead...

adusumillipraveen: https://github.com/hmcts/bulk-scan-shared-infrastructure/pull/144

PR #555, opened 2019-10-30
Deactivate the can-i-deploy step

slyg:
As of now, the pact hooks are not implemented making this check fail the pipelines for consumers.

Once more providers trigger contract verifications, we can activate it.


slyg:
Thx for the review @timja, I'm not authorised to merge this PR though, happy to be added to the right group or for anyone to merge it on my behalf.

PR #575, opened 2019-12-13
Changed variable reference to jenkins subscription env var

gbstan92:
[AB#697](https://dev.azure.com/hmcts/90472b63-dc8d-4ae6-8819-a32bd211e914/_workitems/edit/697)

There was a variable reference set wrongly to use environment subscription instead of the jenkins subscription.

### DIFF
--- a/vars/withSubscription.groovy
+++ b/vars/withSubscription.groovy
@@ -90,7 +90,8 @@ def identityBasedLogin(String subscription, Closure body) {
     withAzureKeyvault([
       [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: "${subscription}-subscription-id", version: '', envVariable: 'ARM_SUBSCRIPTION_ID']
     ]) {
-      az "account set --subscription ${env.JENKINS_SUBSCRIPTION_NAME}"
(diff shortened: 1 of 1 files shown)


vars/withSubscription.groovy
```diff
@@ -90,7 +90,7 @@ def identityBasedLogin(String subscription, Closure body) {
     withAzureKeyvault([
       [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: "${subscription}-subscription-id", version: '', envVariable: 'ARM_SUBSCRIPTION_ID']
     ]) {
-      az "account set --subscription ${env.ARM_SUBSCRIPTION_ID}"
+      az "account set --subscription ${env.JENKINS_SUBSCRIPTION_NAME}"
```
timja: no you should do this on the azJenkins, `az` is for the subscription you're currently working in
```suggestion
      az "account set --subscription ${env.ARM_SUBSCRIPTION_ID}"
      azJenkins "account set --subscription ${env.JENKINS_SUBSCRIPTION_NAME}"
```

They use different AZURE_CONFIG_DIR so that it can have different accounts set

PR #578, opened 2020-01-08
Fix build fixed notifcations

adusumillipraveen:
We don't get fixed notifications when the previous build is aborted

### DIFF
--- a/vars/notifyBuildFixed.groovy
+++ b/vars/notifyBuildFixed.groovy
@@ -30,7 +30,7 @@ def call(Map args = [:]) {
   }
 
   try {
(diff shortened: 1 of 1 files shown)


vars/notifyBuildFixed.groovy
```diff
@@ -30,7 +30,7 @@ def call(Map args = [:]) {
   }
 
   try {
-    if (currentBuild.getPreviousBuild()?.getResult() == 'FAILURE') {
```
timja: tested somewhere?

adusumillipraveen: not yet, but will post results.

PR #588, opened 2020-02-04
Remove deprecated `enableDockerBuild`

timja:


test/uk/gov/hmcts/contino/MockDocker.groovy
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.contino
```
njrich28: contino? i guess because it maps to the Docker class

timja: yeah 😢

PR #590, opened 2020-02-04
SIDM - Merge SIDAM requirements to master

henrydobson:
### Change description ###

Merge SIDAM requirements to master

**Publishing Charts**

> This is likely the most controversial change and I'm happy to discuss/contribute to more IDAM specific steps to achieve the desired result.

SIDAM still deploys to both ASE and AKS. As charts publish only when the `master` branch build [runs](https://github.com/hmcts/cnp-jenkins-library/blob/4a6b7c1fbdef7cd0c081b01680b49d56d3144ede/vars/withPipeline.groovy#L94-L143) and that triggers an ASE deployment to AAT and Prod, which is not desirable. To develop and publish a chart, we would have to merge to master Not desirable.
Resolution: Publish chart in preview.

**idam-\* Environments**

SIDAM uses separate ASE environments to the rest of Reform. For example: idam-preview, idam-aat and idam-prod. To make this possible, the environment names are overridden in the Jenkinsfile ie. `env.PREVIEW_ENVIRONMENT_NAME = 'idam-preview'`. AKS however, does not use separate environments so to make it possible to define the correct environments for both ASE and AKS, we require a check to remove `idam-` from the AKS environment names.

**ASE/AKS Callback Uniqueness**

SIDAM needed to set unique callbacks for AKS and ASE however the callback name was common so we've added a unique callback name.

**Does this PR introduce a breaking change?** (check one with "x")

```
[x] Yes
[ ] No
```


vars/sectionDeployToAKS.groovy
```diff
@@ -52,6 +52,8 @@ def call(params) {
                 deploymentNumber = githubCreateDeployment()
               }
               withAksClient(subscription, environment) {
+                params.environment = params.environment.replace('idam-', '')
+                println "Using AKS environment: ${params.environment}"
```
nikola-naydenov-hmcts: should probably use the logger here

vars/sectionDeployToDeploymentTarget.groovy
```diff
@@ -109,7 +109,7 @@ def call(params) {
           stage("Functional Test - ${environmentDt} (staging slot)") {
             testEnv(deployer.getServiceUrl(environmentDt, "staging"), mergedTfOutput) {
               pcr.callAround("functionalTest:${environmentDt}") {
-                timeoutWithMsg(time: 40, unit: 'MINUTES', action: 'Functional Test') {
+                timeoutWithMsg(time: 120, unit: 'MINUTES', action: 'Functional Test') {
```
nikola-naydenov-hmcts: We dont need this anymore with the new serenity tests and the cleaner AAT

henrydobson:
```suggestion
                timeoutWithMsg(time: 40, unit: 'MINUTES', action: 'Functional Test') {
```

vars/sectionDeployToAKS.groovy
henrydobson:
```suggestion
                log.info("Using AKS environment: ${params.environment}")
```

src/uk/gov/hmcts/contino/AppPipelineConfig.groovy
```diff
@@ -5,6 +5,7 @@ class AppPipelineConfig extends CommonPipelineConfig implements Serializable {
   Map<String, String> vaultEnvironmentOverrides = ['preview':'aat']
   String vaultName
   boolean migrateDb = false
+  boolean uniqueDbMigrateSecretNames = false;
```
henrydobson:
```suggestion
```

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -16,13 +16,14 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.vaultSecrets = vaultSecrets
   }
 
-  void enableDbMigration(String dbMigrationVaultName = "") {
+  void enableDbMigration(String dbMigrationVaultName = "", boolean uniqueSecretNames = false) {
```
henrydobson:
```suggestion
  void enableDbMigration(String dbMigrationVaultName = "") {
```

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -16,13 +16,14 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
...
     if (dbMigrationVaultName == "") {
       WarningCollector.addPipelineWarning("deprecated_enable_db_migration_no_vault)", "enableDbMigration() is deprecated, please use enableDbMigration(<vault-name>)", new Date().parse("dd.MM.yyyy", "05.09.2019"))
     }
 
     config.migrateDb = true
     config.dbMigrationVaultName = dbMigrationVaultName
+    config.uniqueDbMigrateSecretNames = uniqueSecretNames
```
henrydobson:
```suggestion
```

vars/sectionDeployToEnvironment.groovy
```diff
@@ -60,7 +60,7 @@ def call(params) {
 
                   builder.dbMigrate(
                     tfOutput.vaultName ? tfOutput.vaultName.value : "${config.dbMigrationVaultName}-${environment}",
-                    tfOutput.microserviceName? tfOutput.microserviceName.value : component
+                    config.uniqueDbMigrateSecretNames? "${product}-${component}" : component
```
henrydobson:
```suggestion
```

vars/sectionDeployToAKS.groovy
```diff
@@ -81,9 +85,11 @@ def call(params) {
           onFunctionalTestEnvironment(environment) {
             stage("Functional Test - AKS ${environment}") {
               testEnv(aksUrl) {
-                pcr.callAround("functionalTest:${environment}") {
-                  timeoutWithMsg(time: 40, unit: 'MINUTES', action: 'Functional Test - AKS') {
-                    builder.functionalTest()
+                pcr.callAround("functionalTest-aks:${environment}") {
```
timja: why do you need this?

henrydobson: We need to override environment targets and test URLs for AKS but not for ASE. https://github.com/hmcts/idam-api/blob/preview/Jenkinsfile_CNP#L55:L63

timja: but can't you do `smoketest:idam-preview` ? how is it special on AKS and not ASE?

henrydobson: It's a while ago now but I think the reason I put this in place if because with the addition of AKS deployments for PRs and removal of spinInfra, we lost the environment vars that were being set from the terraform output. This meant we needed to set them and so I set them in the callbacks but the variable is different for ASE and AKS so I couldn't use `smoketest:idam-preview` because that set it for both.

timja: but you're only setting this?
https://github.com/hmcts/idam-api/blob/preview/Jenkinsfile_CNP#L58

henrydobson: Yes and also overriding the integration environment with https://github.com/hmcts/idam-api/blob/preview/Jenkinsfile_CNP#L57. We also do similar things in web-public and web-admin. We would not want to set this globally or for ASE as it will break ASE smoke and functional tests.

nikola-naydenov-hmcts: @timja we are planning to be fully over to AKS within the next two months, and at that time we can revert these changes. With the custom branches being blocked we will have no choice but to do this for now

timja: This isn't a blocker, just trying to understand, I don't fully understand why its needed, but dw it doesn't cause any harm

PR #595, opened 2020-02-07
Warn about AAD identity being set for preview

timja:


resources/uk/gov/hmcts/helm/warn-about-aad-identity-preview-hack.sh
```diff
@@ -0,0 +1,17 @@
...
+
+grep --quiet "aadIdentityName" values.preview.template.yaml
+
+if [ $? -gt 0 ]; then
+  echo "AAD Identity name not found, this is good"
+else
+  echo "==============================================================================="
+  echo "=====  Please remove aadIdentityName from values.preview.template.yaml    ====="
+  echo "=====  This is required as we are going to start enforcing                ====="
+  echo "=====  authentication with your pod's identity and not kvcreds            ====="
```
adusumillipraveen:
```suggestion
  echo "=====  authentication with your pod's identity like all other AKS environments and not kvcreds            ====="
```

PR #599, opened 2020-03-05
Replace document db library with node.js library

timja:
The version we are using is the last one that works, the rest have
issues with date serialisation, I couldn't get the cosmosdb library
working in jenkins due to guava missing methods.

Replacing this with node.js due to a huge number of vulnerabilities in the existing ones dependencies
We're currently at the top of the wrong dashboard because of this


src/uk/gov/hmcts/contino/DocumentPublisher.groovy
```diff
@@ -35,19 +34,29 @@ class DocumentPublisher implements Serializable {
...
+    def collection = collectionLink.split('/').last()
 
-    try {
-      documents.each {
-        documentClient.createDocument(collectionLink, new Document(it), null, false)
-      }
-    }
-    finally {
-      documentClient.close()
+    steps.sh "mkdir -p /tmp/metrics-reporting"
```
njrich28: Any issue with concurrency here? If multiple builds are using the same directory at the same time?

src/uk/gov/hmcts/contino/MetricsPublisher.groovy
```diff
@@ -66,19 +65,26 @@ class MetricsPublisher implements Serializable {
...
+      chmod +x metrics-publisher.js
+      yarn install
+
+      export COSMOS_DB_URL=${cosmosDbUrl}
+      export COSMOSDB_TOKEN_KEY=${env.COSMOSDB_TOKEN_KEY}
+      export COSMOS_COLLECTION_ID=pipeline-metrics
 
+      ./metrics-publisher.js '${data}'
+    """
+    steps.echo "Published metrics"
```
njrich28: What's the performance like? This gets run on every stage of every build

src/uk/gov/hmcts/contino/MetricsPublisher.groovy
```diff
@@ -66,19 +65,26 @@ class MetricsPublisher implements Serializable {
     ]
   }
 
-  @NonCPS
   private def createDocument(metrics, cosmosDbUrl) {
-    def client = new DocumentClient(cosmosDbUrl, env.COSMOSDB_TOKEN_KEY, null, null)
+    steps.sh "mkdir -p /tmp/metrics-reporting"
```
njrich28: As above

test/uk/gov/hmcts/pipeline/deprecation/WarningCollectorTest.groovy
```diff
@@ -57,13 +57,13 @@ class WarningCollectorTest extends Specification {
     assertThat(message).isEqualTo(nextDayFormattedDate+" ( tomorrow )")
   }
 
-  def "getMessageByDays() with a week later should return in 6 days"() {
+  def "getMessageByDays() with a week later should return in 7 days"() {
```
njrich28: Ah good! This always seemed wrong to me. The tests were a bit flakey before in that sometimes they'd return 7 and sometimes return 6 - presume that's fixed now?

src/uk/gov/hmcts/contino/DocumentPublisher.groovy
timja: we only run one executor so it's fine

src/uk/gov/hmcts/contino/MetricsPublisher.groovy
timja: takes one second or less,
https://sandbox-build-beta.platform.hmcts.net/job/timj-test/17/console

test/uk/gov/hmcts/pipeline/deprecation/WarningCollectorTest.groovy
timja: I think so, I changed it to take the time at the start of the day

PR #603, opened 2020-04-02
Allow overriding mgmtstatestore name

timja:


njrich28:
The test pipeline has actually passed - but it's not reflected above. https://sandbox-build-beta.platform.hmcts.net/blue/organizations/jenkins/HMCTS_Sandbox_Pipeline_Test%2Fcnp-jenkins-library/detail/PR-603/3/pipeline/

PR #614, opened 2020-05-18
Publish OWASP vulnerability reports to CosmosDB

banderous:
Send details of vulnerable dependencies to CosmosDB for reporting

Uses the OWASP dependency checker's JSON report format to extract a list of dependencies with known vulnerabilities, and sends to Cosmos along with some build information.

Went with modifying the Gradle builder directly since this is only applicable to the Gradle builder and not a more generic pipeline piece that could be composed elsewhere.


src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -88,10 +100,56 @@ class GradleBuilder extends AbstractBuilder {
...
+        git_url                      : env.GIT_URL,
+      ],
+      report: report
+    ]
+    return JsonOutput.toJson(result)
+  }
+
+  @NonCPS
+  private def createDocument(String reportJSON) {
+    def client = new DocumentClient(cosmosDbUrl, steps.env.COSMOSDB_TOKEN_KEY, null, null)
```
timja, line 144: isn't this autocloseable?
https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.documentdb.documentclient?view=azure-java-stable (doesn't map to the version we're using so it might not)

banderous: withCloseable for AutoCloseable is since Groovy 2.5.0 - http://docs.groovy-lang.org/docs/latest/html/groovy-jdk/java/lang/AutoCloseable.html

PR #625, opened 2020-05-22
Helm 3 doesnt have requirements.yaml

adusumillipraveen:
Fails if used with helm 3 charts.

### DIFF
--- a/resources/uk/gov/hmcts/helm/check-helm-version-bumped.sh
+++ b/resources/uk/gov/hmcts/helm/check-helm-version-bumped.sh
@@ -13,6 +13,15 @@ else
   DIFF_IN_VALUES=false
 fi
 
+git diff -s --exit-code origin/master charts/"${CHART_DIRECTORY}"/Chart.yaml
+
+if [ $? -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


resources/uk/gov/hmcts/helm/check-helm-version-bumped.sh
```diff
@@ -13,13 +13,14 @@ else
...
 
-git diff -s --exit-code origin/master charts/"${CHART_DIRECTORY}"/requirements.yaml
-
-if [ $? -eq 1 ]; then
-  echo "Diff in requirements.yaml detected"
-  DIFF_IN_REQUIREMENTS=true
-else
-  DIFF_IN_REQUIREMENTS=false
+DIFF_IN_REQUIREMENTS=false
+if [[ -f charts/"${CHART_DIRECTORY}"/requirements.yaml ]]
```
timja, line 17: then Chart.yaml needs checking in this case?

adusumillipraveen: Thanks, addressed.

PR #629, opened 2020-06-05
Bump chart version before failing the build

adusumillipraveen:
Notes:
* Bumps patch version of chart before failing build
* Uses Github App authentication used for code checkout to push commit
* Sets  GIT_CREDENTIALS_ID and GIT_APP_EMAIL_ID while checkout step to be reused in other steps.

### DIFF
--- a/vars/enforceChartVersionBumped.groovy
+++ b/vars/enforceChartVersionBumped.groovy
@@ -2,15 +2,16 @@ def call(Map<String, String> params) {
   def product = params.product
   def component = params.component
   def branch = env.CHANGE_BRANCH
-  def credentialsId= env.GIT_CREDENTIALS_ID
+  def credentialsId = env.GIT_CREDENTIALS_ID
   def gitEmailId = env.GIT_APP_EMAIL_ID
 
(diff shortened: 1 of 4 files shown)


vars/enforceChartVersionBumped.groovy
```diff
@@ -1,14 +1,19 @@
...
 
   writeFile file: 'check-helm-version-bumped.sh', text: libraryResource('uk/gov/hmcts/helm/check-helm-version-bumped.sh')
 
-  sh """
+  withCredentials([usernamePassword(credentialsId: credentialsId, passwordVariable: 'BEARER_TOKEN', usernameVariable: 'APP_ID')]) {
+    def bearerToken = env.BEARER_TOKEN
+    sh """
     chmod +x check-helm-version-bumped.sh
-    ./check-helm-version-bumped.sh $product $component
+    ./check-helm-version-bumped.sh $product $component $branch $credentialsId $bearerToken $gitEmailId
```
timja, line 14: why not use the env variables that it's setting?

BEARER_TOKEN and APP_ID?

adusumillipraveen: It fails tests that $BEARER_TOKEN is not defined , was just working around it :)

PR #653, opened 2020-07-14
Scan only runtime dependencies with OWSAP dependency checker

banderous:
Many projects are running with default OWASP checker settings which means a lot of non runtime configurations are scanned and a lot of noise is generated.

I have also directly configured the checker to fail the build if a CVE is found, rather than passing a flag to the project build script and relying on the project to respect it.

Related PR for the RSE Gradle plugin: https://github.com/hmcts/rse-gradle-java-plugin/pull/29/files

### DIFF
--- a/resources/uk/gov/hmcts/gradle/init.gradle
+++ b/resources/uk/gov/hmcts/gradle/init.gradle
@@ -41,19 +41,22 @@ settingsEvaluated { settings ->
 allprojects {
   afterEvaluate { project ->
     if (project.hasProperty('dependencyCheck')) {
-      // Any CVE should fail the build.
-      project.dependencyCheck.failBuildOnCVSS = 0
+      configure(project) {
+        dependencyCheck {
+          // Any CVE should fail the build.
+          failBuildOnCVSS = 0
 
-      // JSON dependency check reports are logged to cosmos for reporting.
-      project.dependencyCheck.formats += 'json'
+          // JSON dependency check reports are logged to cosmos for reporting.
+          formats += 'json'
 
-      // Scan only the dependencies used at application runtime.
-      // See https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management
-      project.dependencyCheck.scanConfigurations += 'runtimeClasspath'
-      // Scan and skip configurations are mutually exclusive,
-      // so we must clear any skipped configurations the build may have configured.
-      project.dependencyCheck.skipConfigurations = []
+          // Scan only the dependencies used at application runtime.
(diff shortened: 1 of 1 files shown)


resources/uk/gov/hmcts/gradle/init.gradle
```diff
@@ -38,11 +38,21 @@ settingsEvaluated { settings ->
   }
 }
 
-// Enable JSON dependency check reports for monitoring of CVEs and suppressions.
 allprojects {
   afterEvaluate { project ->
     if (project.hasProperty('dependencyCheck')) {
+      // Any CVE should fail the build.
+      project.dependencyCheck.failBuildOnCVSS = 0
```
timja, line 45: can't this all be in a closure? untested

```suggestion
      project {
        dependencyCheck {
          failBuildOnCVSS = 0
       	  formats += json
       	}
      }
```

resources/uk/gov/hmcts/gradle/init.gradle
```diff
@@ -38,11 +38,21 @@ settingsEvaluated { settings ->
...
+
+      // JSON dependency check reports are logged to cosmos for reporting.
       project.dependencyCheck.formats += 'json'
+
+      // Scan only the dependencies used at application runtime.
+      // See https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_plugin_and_dependency_management
+      project.dependencyCheck.scanConfigurations += 'runtimeClasspath'
```
timja, line 52: this would be more obvious to me, I had to go check the docs to understand why appending one worked, ([] implies all configuration).

```suggestion
      project.dependencyCheck.scanConfigurations = ['runtimeClasspath']
```

banderous: I made it additive so that any existing configurations are respected and not cleared out.

banderous: (since this is executing after build configuration)

timja: is it an issue if someone already has `runtimeClasspath` in their `scanConfigurations`?

resources/uk/gov/hmcts/gradle/init.gradle
banderous: 👍

resources/uk/gov/hmcts/gradle/init.gradle
banderous: The dependency checker just checks that list for existence so duplicates should be no problem. I've added a check nevertheless.

PR #654, opened 2020-07-14
Refactor cve reporting

timja:
This will allow code reuse for node.js

### DIFF
--- a/test/uk/gov/hmcts/pipeline/CVEPublisherTest.groovy
+++ b/test/uk/gov/hmcts/pipeline/CVEPublisherTest.groovy
@@ -1,0 +1,36 @@
+package uk.gov.hmcts.pipeline
+
+import spock.lang.Specification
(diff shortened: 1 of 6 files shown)


test/uk/gov/hmcts/pipeline/CVEPublisherTest.groovy
```diff
@@ -0,0 +1,49 @@
...
+      ]
+    ]
+
+    when:
+    cvePublisher.publishCVEReport(report)
+
+    then:
+    notThrown(Exception.class)
+
+    1 * documentClient.createDocument('dbs/jenkins/colls/cve-reports', _, null, false)
```
timja, line 46: I tried for far too long to verify that the report is correct but it doesn't work,
I think it's because the `Document` class doesn't implement equals.

could extract something else out but seems overkill

PR #656, opened 2020-07-15
Add reporting for CVE issues on node.js

timja:
AB#994


src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -102,9 +105,45 @@ class YarnBuilder extends AbstractBuilder {
     """
     } catch(ignored) { // TODO remove try catch after pipeline warning expires
       WarningCollector.addPipelineWarning("node_cve", "CVEs found for Node.JS, update your dependencies / ignore false positives", new Date().parse("dd.MM.yyyy", "28.07.2020"))
+    } finally {
+      String issues = steps.readFile('yarn-audit-issues')
```
banderous, line 109: If the reporting itself breaks do we want to block the pipeline? ie. do we want some exception handling in this finally?  Could we log any exceptions here such that we get notified and can investigate?

timja: discussed on slack, fine without it for now imo,

We want to know if report preparation fails.

The actual publishing if it fails won't break the build

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -102,9 +105,68 @@ class YarnBuilder extends AbstractBuilder {
...
+        .publishCVEReport('node', cveReport)
     }
   }
 
+  def prepareCVEReport(String issues, String knownIssues) {
+    def jsonSlurper = new JsonSlurper()
+    List<Object> issuesParsed = issues.split( '\n' ).collect { jsonSlurper.parseText(it) }
+
+    Object summary = issuesParsed.find { it.type == 'auditSummary' }
+    issuesParsed.removeIf { it.type == 'auditSummary' }
```
banderous, line 127: Do we need this removal now?

PR #657, opened 2020-07-23
Add .dockerignore file

luigibk:

Notes:
*
*
*

### DIFF
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -101,6 +101,7 @@ def call(params) {
 
             pcr.callAround('dockerbuild') {
               timeoutWithMsg(time: 30, unit: 'MINUTES', action: 'Docker build') {
+                writeFile file: '.dockerignore', text: libraryResource('uk/gov/hmcts/.dockerignore_build')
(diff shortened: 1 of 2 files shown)


vars/sectionBuildAndTest.groovy
```diff
@@ -101,6 +101,7 @@ def call(params) {
 
             pcr.callAround('dockerbuild') {
               timeoutWithMsg(time: 30, unit: 'MINUTES', action: 'Docker build') {
+                writeFile file: '.dockerignore', text: libraryResource('uk/gov/hmcts/.dockerignore_build')
```
timja, line 104: what if the docker ignore file already exists?

luigibk: Where? I looked for one.

timja: we never got a reliable fix, we turned off the files that were causing the issue in some builds which helped but sometimes people still hit it.

if this doesn't work I would either stop removing the file after use, or only remove it after the parallel stages have finished

luigibk:
> worth a try

Why is that file deleted anyway?

timja:
> > worth a try
> 
> Why is that file deleted anyway?

just because it was a file being created in the workspace, could just not delete it

PR #665, opened 2020-08-04
Setting team config to env vars

adusumillipraveen:
AB#1038

Notes:
* We are calling TeamConfig at many places , sometimes for the same thing.
* This will set all the team config as env variables, so we can just access environment variables instead.

### DIFF
--- a/vars/dockerAgentSetup.groovy
+++ b/vars/dockerAgentSetup.groovy
@@ -1,9 +1,7 @@
-import uk.gov.hmcts.pipeline.TeamConfig
 import uk.gov.hmcts.contino.azure.KeyVault
 
 def call(String product) {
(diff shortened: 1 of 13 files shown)


vars/dockerAgentSetup.groovy
```diff
@@ -1,9 +1,7 @@
-import uk.gov.hmcts.pipeline.TeamConfig
 import uk.gov.hmcts.contino.azure.KeyVault
 
 def call(String product) {
-  def teamConfig = new TeamConfig(this)
-  if (teamConfig.isDockerBuildAgent(product)) {
+  if (env.IS_DOCKER_BUILD_AGENT) {
```
timja, line 4: make sure you test this as env vars are strings

PR #668, opened 2020-08-05
Add option to override container registry 

adusumillipraveen:
Notes:
* Doesnt override helm registry as we still want them to publish to `hmctspublic`

### DIFF
--- a/vars/sectionPromoteBuildToStage.groovy
+++ b/vars/sectionPromoteBuildToStage.groovy
@@ -37,7 +37,7 @@ def call(params) {
 
   stageWithAgent("${deploymentStage.label} build promotion", product) {
     withAcrClient(subscription, product) {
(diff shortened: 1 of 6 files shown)


vars/sectionPromoteBuildToStage.groovy
```diff
@@ -37,9 +37,9 @@ def call(params) {
 
   stageWithAgent("${deploymentStage.label} build promotion", product) {
     withAcrClient(subscription, product) {
-
+        def imageRegistry = env.TEAM_CONTAINER_REGISTRY ? env.TEAM_CONTAINER_REGISTRY : env.REGISTRY_NAME
```
timja, line 40: simpler?
```suggestion
        def imageRegistry = env.TEAM_CONTAINER_REGISTRY ?: env.REGISTRY_NAME
```

PR #670, opened 2020-08-07
Avoid duplicate az login(s) + ssh key downloads from vault

luigibk:

[AB#1049]

Notes:
*
*
*


timja:
the original approach should work too, and would give everyone benefits i would think

luigibk:
> the original approach should work too, and would give everyone benefits i would think

I thought so initially ... but I am having second thoughts as it's riskier

luigibk:
> The files class will likely only run on the master

fileExists is not executing in the right docker container so I gave it a try, but yeah that did not help either

PR #697, opened 2020-10-12
Point artifactory in sandbox to new domain

timja:
As part of tactical decommissioning we need to remove the proxyin server that this service is using,
domain name needs to change as the ingress controller is only setup with .sandbox.platform.hmcts.net atm and it makes sense to change it

### DIFF
--- a/Jenkinsfile_pipeline_test
+++ b/Jenkinsfile_pipeline_test
@@ -1,5 +1,5 @@
 #!groovy
-@Library('Infrastructure') _
+@Library('Infrastructure@repoint-sandbox-artifactory') _
 import uk.gov.hmcts.contino.GradleBuilder
 
 def channel = '#rpe-build-notices'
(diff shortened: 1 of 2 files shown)


timja:
Need to debug locally, will do tomorrow:

> sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

Probably an issue with the cert chain

naikajah (dismissed):

timja:
works in browser FTR

timja:
I've added the chain in https://github.com/hmcts/cnp-flux-config/pull/6424, curl now works from a linux VM, previously it complained about the issuer.

It worked on my mac for everything without the change 👅 , browser, curl, java...

Jenkinsfile_pipeline_test
```diff
@@ -1,5 +1,5 @@
 #!groovy
-@Library('Infrastructure') _
+@Library('Infrastructure@repoint-sandbox-artifactory') _
```
timja, line 2:
```suggestion
@Library('Infrastructure') _
```

PR #704, opened 2020-10-28
Use default installation for jdk

adusumillipraveen:

### DIFF
--- a/src/uk/gov/hmcts/contino/AbstractBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/AbstractBuilder.groovy
@@ -36,8 +36,8 @@ abstract class AbstractBuilder implements Builder, Serializable {
   def setupToolVersion() {
 
     //Setting Up JAVA_HOME here as its used for Sonar scan even in YarnBuilder.
-    if (steps.fileExists("/usr/share/jdk-11.0.2")) {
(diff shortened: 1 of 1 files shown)


src/uk/gov/hmcts/contino/AbstractBuilder.groovy
```diff
@@ -36,8 +36,8 @@ abstract class AbstractBuilder implements Builder, Serializable {
   def setupToolVersion() {
 
     //Setting Up JAVA_HOME here as its used for Sonar scan even in YarnBuilder.
-    if (steps.fileExists("/usr/share/jdk-11.0.2")) {
-      steps.env.JAVA_HOME = "/usr/share/jdk-11.0.2"
+    if (steps.fileExists("/usr/lib/jvm/java-11-openjdk")) {
```
timja, line 39: Is this even needed when there’s only 1 installed?

timja (approved):

src/uk/gov/hmcts/contino/AbstractBuilder.groovy
adusumillipraveen: There is still java 8 on the VM, couldn’t trace it down.

adusumillipraveen: Merging it as would like to test couple of builds before others start using Jenkins.

PR #714, opened 2020-12-15
Add parallel crossBrowserTest() to run nightly crossbrowser tests in parallel steps

Rob-Doherty:
This change adds a new parallel crossBrowserTest() feature to run nightly crossbrowser tests in parallel Jenkins steps. The feature can be used by adding the browser groups that are to run in parallel in the service's Jenkinsfile_nightly (i.e. extending the existing enableCrossBrowserTest() call), like so:
```
enableCrossBrowserTest(['chrome', 'firefox', 'safari', 'microsoft'])
```

**Notes:**
* The new step can be overridden to run any browser / browser group command that the service requires

### DIFF
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -42,7 +42,6 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
 
   void enableParallelCrossBrowserTest(List<String> browsers = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120) {
     config.crossBrowserTestTimeout = timeout
-    config.parallelCrossBrowserTest = true
     config.parallelCrossBrowsers = browsers
   }
 
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -40,6 +40,12 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.crossBrowserTest = true
   }
 
+  void enableParallelCrossBrowserTest(List<String> browsers = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120) {
+    config.crossBrowserTestTimeout = timeout
+    config.parallelCrossBrowserTest = true
+    config.parallelCrossBrowsers = browsers
+  }
+
   void enableSecurityScan(int timeout = 120) {
     config.securityScanTimeout = timeout
     config.securityScan = true
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -40,7 +40,7 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.crossBrowserTest = true
   }
 
-  void enableParallelCrossBrowserTest(List<String> browsers = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120) {
+  void enableCrossBrowserTest(List<String> browsers, int timeout = 120) {
     config.crossBrowserTestTimeout = timeout
     config.parallelCrossBrowsers = browsers
   }
--- a/src/uk/gov/hmcts/contino/Builder.groovy
+++ b/src/uk/gov/hmcts/contino/Builder.groovy
@@ -9,7 +9,7 @@ interface Builder {
   def performanceTest()
   def apiGatewayTest()
   def crossBrowserTest()
-  def parallelCrossBrowserTest(String browser)
+  def crossBrowserTest(String browser)
   def securityCheck()
   def addVersionInfo()
   def mutationTest()
--- a/src/uk/gov/hmcts/contino/Builder.groovy
+++ b/src/uk/gov/hmcts/contino/Builder.groovy
@@ -9,7 +9,6 @@ interface Builder {
   def performanceTest()
   def apiGatewayTest()
   def crossBrowserTest()
-  def crossBrowserTest(String browser)
   def securityCheck()
   def addVersionInfo()
   def mutationTest()
--- a/src/uk/gov/hmcts/contino/Builder.groovy
+++ b/src/uk/gov/hmcts/contino/Builder.groovy
@@ -9,6 +9,7 @@ interface Builder {
   def performanceTest()
   def apiGatewayTest()
   def crossBrowserTest()
+  def parallelCrossBrowserTest(String browser)
   def securityCheck()
   def addVersionInfo()
   def mutationTest()
--- a/vars/sectionNightlyTests.groovy
+++ b/vars/sectionNightlyTests.groovy
@@ -50,50 +50,62 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
       }
     }
     if (config.parallelCrossBrowserTest) {
-      stageWithAgent("Cross browser tests", product) {
-        warnError('Failure in parallelCrossBrowserTest') {
-          Set<String> browsers = config.parallelCrossBrowsers.collect{ it.toLowerCase() }.toSet()
-          parallel(
-//            browsers.each {browser ->
-//              browser : {
-//                pcr.callAround('parallelCrossBrowserTest') {
-//                  timeoutWithMsg(time: config.crossBrowserTestTimeout, unit: 'MINUTES', action: 'Cross browser test') {
-//                    builder.parallelCrossBrowserTest(it)
-//                  }
-//                }
+      Set<String> browsers = config.parallelCrossBrowsers.collect{ it.toLowerCase() }.toSet()
+      parallel(
+//        browsers.each {browser ->
+//          browser : {
+//            pcr.callAround('parallelCrossBrowserTest') {
+//              timeoutWithMsg(time: config.crossBrowserTestTimeout, unit: 'MINUTES', action: 'Cross browser test') {
+//                builder.parallelCrossBrowserTest(it)
 //              }
 //            }
-            "Chrome" : {
+//          }
+//        }
+        "Chrome": {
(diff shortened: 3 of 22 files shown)


vars/sectionNightlyTests.groovy
```diff
@@ -49,6 +49,52 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
         }
       }
     }
+    if (config.parallelCrossBrowserTest) {
+      stageWithAgent("Cross browser tests", product) {
+        warnError('Failure in parallelCrossBrowserTest') {
+          Set<String> browsers = config.parallelCrossBrowsers.collect{ it.toLowerCase() }.toSet()
+          parallel(
+//            browsers.each {browser ->
```
timja, line 57: remove commented out code if you aren't using it

vars/sectionNightlyTests.groovy
```diff
@@ -49,6 +49,52 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
         }
       }
     }
+    if (config.parallelCrossBrowserTest) {
+      stageWithAgent("Cross browser tests", product) {
+        warnError('Failure in parallelCrossBrowserTest') {
+          Set<String> browsers = config.parallelCrossBrowsers.collect{ it.toLowerCase() }.toSet()
```
timja, line 55: not in use?

vars/sectionNightlyTests.groovy
```diff
@@ -49,6 +49,52 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
         }
       }
     }
+    if (config.parallelCrossBrowserTest) {
```
timja, line 52: why not just re-use the existing one? best to give everyone the benefit?

probably worth the extra testing?

any blockers for using the existing one?

timja (commented):
few comments

vars/sectionNightlyTests.groovy
Rob-Doherty: I'm hoping to test 2 ways of running these in parallel - the current one where the browsers being called are fixed,and then the commented out way where the services will be able to override the default browsers with their own requirements (but I'm not sure i have this correct yet)

I'll definitely remove whichever one doesn't make the cut 👍

vars/sectionNightlyTests.groovy
Rob-Doherty: yeah, I'm hoping to use it if I can get the override functionality working properly

vars/sectionNightlyTests.groovy
Rob-Doherty: I don't think all the teams have implemented the crossbrowser tests in the exact same way, and I'm pretty sure they don't have separate scripts in package.json to support
```
yarn test:crossbrowser-chrome
yarn test:crossbrowser-firefox
...
```
But that's an interesting thought, as a few teams already support using a `BROWSER_GROUP` environment variable to define the browser to run, so maybe that's a better route to go 🤔. I'll have a look at a few services to see how common that approach is and maybe move towards that instead

Rob-Doherty: But even still, I don't think we'd be able to update the current functionality without breaking alot of service's current tests.

I can add a `@deprecated` flag to the old method once I've tested the new one fully

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -40,6 +40,12 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.crossBrowserTest = true
   }
 
+  void enableParallelCrossBrowserTest(List<String> browsers = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120) {
```
adusumillipraveen, line 43: A general suggestion, why not have it as an overloaded method with additional parameter like parallelBrowsers and change code to behave based on this ? That might remove some duplicate lines.

vars/sectionNightlyTests.groovy
timja: sure, a standard `BROWSER` env flag would be good

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
Rob-Doherty: that's a really interesting idea @adusumillipraveen , I'll look into that for the actual `parallelCrossBrowserTest()` method which could definitely be renamed to overload the existing `crossBrowserTest()` method. 

But I think we'd still want a separate `enableParallelCrossBrowserTest()` here because I'm hoping most services will end up using the default parameters like 
```
  enableParallelCrossBrowserTest()
```
 so unless I'm misunderstanding we'd have to supply something to differentiate between the similar named methods like a 'parallel: true' value or something, which I'm not sure gives us much benefit over a different method name

Rob-Doherty:
This change isn't stale - it's just been in testing

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
Rob-Doherty: done

vars/sectionNightlyTests.groovy
Rob-Doherty: I've made those changes now, using the `BROWSER_GROUP` env var to set the browser type, as this is already a commonly used variable for crossbrowser testing in service teams. It also reduces any potential impact from an env var called BROWSER already being used for something else.

Rob-Doherty:
Thanks for your earlier reviews @timja and @adusumillipraveen, I've incorporated your ideas and been through many rounds of testing, and the current changes are working well. Would you mind re-reviewing?

It'd be great to get this merged in soon to benefit the work I'm doing for Probate, and then to roll it out to the other services.

**Testing examples:**

Screenshot 2021-01-07 at 09 00 43

parallel_jenkins_nightly_crossbrowser_pipeline_probate

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -40,6 +40,12 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
...
+    config.crossBrowserTestTimeout = timeout
+    config.parallelCrossBrowserTest = true
```
adusumillipraveen, line 45: this property looks redundant to me. If parallelCrossBrowsers are set, doesn't it mean we are asking for parallel execution.  Also, can we rename this method to  something like. 

`enableCrossBrowserTest(List<String> parallelBrowsersList = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120)`

src/uk/gov/hmcts/contino/Builder.groovy
```diff
@@ -9,6 +9,7 @@ interface Builder {
   def performanceTest()
   def apiGatewayTest()
   def crossBrowserTest()
+  def crossBrowserTest(String browser)
```
adusumillipraveen, line 12: cant this be the only method and call it with empty string to keep the behaviour of   `def crossBrowserTest()` intact?

adusumillipraveen (commented):
some minor comments.

src/uk/gov/hmcts/contino/Builder.groovy
Rob-Doherty: good call - will change this

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
Rob-Doherty: Good call about the redundant `config.parallelCrossBrowserTest` - I'll remove it.

Rob-Doherty: But I might not have explained myself very well in your previous question about overloading this method. Unless I'm misunderstanding your point, I believe your method rename would not allow services to use a default browser list. This would require each service to configure that themselves even though there's likely be a lot of repetition, which is all I'm trying to avoid. It would also need a bit more understanding from each team as to how to configure it correctly, rather than being a bit more plug-and-play like I'm hoping.

So this scenario with your suggested rename wouldn't be valid, as it won't know which method to use:
```
void enableCrossBrowserTest(int timeout = 120) { 
... 
}

void enableCrossBrowserTest(List<String> parallelBrowsersList = ['chrome', 'firefox', 'safari', 'microsoft'], int timeout = 120) { 
...
}
```
```
enableCrossBrowserTest()
```

adusumillipraveen: No, i meant we have only one method which is may be so that it doesn't break existing scripts and we provide this behaviour to everyone. Unless u see a challenge in making parallel execution a problem to all repos.

`void enableCrossBrowserTest(int timeout = 120, List<String> parallelBrowsersList = ['chrome', 'firefox', 'safari', 'microsoft']) {
`

adusumillipraveen (dismissed):

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
Rob-Doherty: (just for the benefit of this thread) 
as we agreed @adusumillipraveen I'll leave the PR as it is, start getting some benefit from the changes, and in the meantime I’ll still look to get other QA's thoughts and create a further PR to either deprecate the old way or overload the existing method as you've suggested here

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -40,6 +40,11 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.crossBrowserTest = true
   }
 
+  void enableCrossBrowserTest(List<String> browsers, int timeout = 120) {
```
adusumillipraveen, line 43: Do u still need this method ? https://github.com/hmcts/cnp-jenkins-library/pull/714/files#diff-08ec10cb390079edcd7ae93cb8d29b85e267a1ed73f3e16c3248af11ce5855efR38

adusumillipraveen (commented):
Minor suggestions

adusumillipraveen (approved):

PR #742, opened 2021-05-14
Allow for preview branch to be tagged as preview-commit

nikola-naydenov-hmcts:
Preview will be tagged only with branch name as the tag.
This change allows it to be re-tagged to also contain the commit tag, i.e preview-ab56cf22

### DIFF
--- a/src/uk/gov/hmcts/contino/azure/Acr.groovy
+++ b/src/uk/gov/hmcts/contino/azure/Acr.groovy
@@ -128,6 +128,7 @@ class Acr extends Az {
     def additionalTag = dockerImage.getShortName(stage)
     def baseTag = (stage == DockerImage.DeploymentStage.PR  || dockerImage.imageTag == 'staging') ? dockerImage.getBaseTaggedName() : dockerImage.getTaggedName()
     if (stage == DockerImage.DeploymentStage.PREVIEW) {
+      // Non-master builds will only have the base branch name as the tag
       baseTag = dockerImage.getBaseShortName();
     }
     this.az "acr import --force -n ${registryName} -g ${resourceGroup} --subscription ${registrySubscription} --source ${baseTag} -t ${additionalTag}"?.trim()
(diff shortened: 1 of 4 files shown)


src/uk/gov/hmcts/contino/azure/Acr.groovy
```diff
@@ -127,6 +127,10 @@ class Acr extends Az {
   def retagForStage(DockerImage.DeploymentStage stage, DockerImage dockerImage) {
     def additionalTag = dockerImage.getShortName(stage)
     def baseTag = (stage == DockerImage.DeploymentStage.PR  || dockerImage.imageTag == 'staging') ? dockerImage.getBaseTaggedName() : dockerImage.getTaggedName()
+    if (stage == DockerImage.DeploymentStage.PREVIEW) {
+      // Non-master builds will only have the base branch name as the tag
+      baseTag = dockerImage.getBaseShortName();
```
timja, line 132: this doesn't look like it will add the commit? is that what you are after?

really struggling to understand this code though, if you say it works I'll approve it =/

nikola-naydenov-hmcts: yeah so I read it like 20 times, and this is what I think is going on..

here baseTag is the source, additionTag is what you want to tag to.. 

getShortName gives you "${imageTag}-${this.commit}" (why is that short?! who knows)
but my images are being tagged only with imageTag (getBaseShortName) by acr.build()

acr build will tag differently if its staging and currently on master the first build is a staging build, which then becomes the prod image and also retagForStage gets called 3 times when going from AAT to Prod, one of the times the 'latest' image tag gets re-tagged as prod..(I think)

I'm.. honestly I'm like 80% sure this will work

nikola-naydenov-hmcts: also here the baseTag will only be different from the additionalTag if its PR, staging or latest, otherwise it seems it will resolve both as the same thing
this code is like a house of cards, it works because everything has been jiggled into the right spot
to do it right I think I have to rewrite the entire tagging process

timja (approved):

PR #743, opened 2021-05-17
Add registry prefix to stage Preview build promotion

nikola-naydenov-hmcts:
I really thought I had it right last time, but I used the wrong method as I misunderstood the getBaseTaggedName() method.

Here is the call that ran with the current code:
```bash
az acr import --force -n hmctspublic -g rpe-acr-prod-rg --subscription DCD-CNP-PROD --source idam/web-admin:preview -t idam/web-admin:preview-200d6edf
```
The error that came back was:
```
ERROR: Login server of the registry is not valid because it is not a fully qualified domain name.
```
I thought I needed baseShortName but baseTaggedName is baseShortName with the registry prefixed, obviously.

Failed build:
https://build.platform.hmcts.net/job/HMCTS_IDAM/job/idam-web-admin/job/preview/55/

### DIFF
--- a/src/uk/gov/hmcts/contino/DockerImage.groovy
+++ b/src/uk/gov/hmcts/contino/DockerImage.groovy
@@ -11,6 +11,7 @@ class DockerImage {
     PR('pr'),
     STAGING('staging'),
     AAT('aat'),
+    PREVIEW('preview'),
     PROD('prod'),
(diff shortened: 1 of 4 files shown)


timja (approved):
this is what i thought it would be last time :p 😄

PR #748, opened 2021-06-08
Add parallel crossbrowser testing for Java projects

Rob-Doherty:
This change extends the recent [parallel crossBrowserTest() feature](https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/sectionNightlyTests.groovy#L53-L70) to run nightly crossbrowser tests in parallel Jenkins steps for Java projects. The feature can be used by adding the browser groups that are to run in parallel in the service's Jenkinsfile_nightly (i.e. extending the existing enableCrossBrowserTest() call), like so:
```
enableCrossBrowserTest(['chrome', 'firefox', 'safari', 'microsoft'])
```

The existing functionality takes this array of strings and makes individual calls to the builder for each string
```
      browsers.each { browser ->
        crossBrowserStages.put(browser.capitalize(), {
          warnError('Failure in crossBrowserTest') {
            pcr.callAround('crossBrowserTest') {
              timeoutWithMsg(time: config.crossBrowserTestTimeout, unit: 'MINUTES', action: 'Cross browser test') {
                builder.crossBrowserTest(browser)
              }
            }
          }
        })
```

However, until this change, there was only a compatible method in the Yarn builder. Now this will work with Java projects too.

**Notes:**
* The new step can be overridden to run any browser / browser group command that the service requires

### DIFF
--- a/src/uk/gov/hmcts/contino/GradleBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/GradleBuilder.groovy
@@ -68,12 +68,12 @@ class GradleBuilder extends AbstractBuilder {
     }
   }
 
-  def crossBrowserTest() {
+  def crossBrowserTest(String browser = "") {
     try {
       // By default Gradle will skip task execution if it's already been run (is 'up to date').
       // --rerun-tasks ensures that subsequent calls to tests against different slots are executed.
       steps.withSauceConnect("reform_tunnel") {
-        gradle("--rerun-tasks crossbrowser")
+        gradle("--rerun-tasks crossbrowser", "BROWSER_GROUP=$browser")
       }
     } finally {
       steps.archiveArtifacts allowEmptyArchive: true, artifacts: 'functional-output/**/*'
@@ -150,9 +150,9 @@ EOF
   }
 
 
-  def gradle(String task) {
+  def gradle(String task, String prepend = "") {
     addInitScript()
-    steps.sh("./gradlew --no-daemon --init-script init.gradle ${task}")
+    steps.sh("${prepend} ./gradlew --no-daemon --init-script init.gradle ${task}")
   }
 
   private String gradleWithOutput(String task) {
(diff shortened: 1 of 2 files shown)


satyachundur (approved):
LGTM

src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -68,12 +68,12 @@ class GradleBuilder extends AbstractBuilder {
     }
   }
 
-  def crossBrowserTest() {
+  def crossBrowserTest(String browser = "") {
```
timja, line 71: this doesn't seem to match your description, you've put an array in the example:
```
enableCrossBrowserTest(['chrome', 'firefox', 'safari', 'microsoft'])
```

but this only accepts a string and you're only testing a string

Rob-Doherty: Good question.
So this is an extension of [existing functionality](https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/sectionNightlyTests.groovy#L53-L70):
```
      browsers.each { browser ->
        crossBrowserStages.put(browser.capitalize(), {
          warnError('Failure in crossBrowserTest') {
            pcr.callAround('crossBrowserTest') {
              timeoutWithMsg(time: config.crossBrowserTestTimeout, unit: 'MINUTES', action: 'Cross browser test') {
                builder.crossBrowserTest(browser)
              }
            }
          }
        })
```
which takes that defined array of strings and makes individual calls to the builder for each string. 
However, until this change, there was only a compatible method in the Yarn builder.

I'll add this detail to the description to make it clearer

timja: I would recommend making this implementation consistent with the YarnBuilder one, duplicating the method rather than using the groovy default parameter, it would end up with less code / clearer what's going on imo.

If you feel strongly it's not clearer then I would ask that you adjust YarnBuilder to be consistent, thanks!

Rob-Doherty: happy to follow the Yarn format if it improves readability, even if it increases the code slightly. I'll do that now

Rob-Doherty: done

timja (approved):

PR #755, opened 2021-06-24
Implementation of fortify repository scan to NightlyPipeline

kaanaktas:
Fortify integration with application pipelines

Notes:
* Added fortifyScan new stage 
* Implemented new stage in Nightly Pipeline

Tested projects:
* Java: https://build.platform.hmcts.net/view/CDM/job/HMCTS_Nightly_CDM/job/ccd-user-profile-api/job/nightly-fod-scan-ccd-1257/
* Angular: https://build.platform.hmcts.net/view/CDM/job/HMCTS_Nightly_CDM/job/ccd-case-management-web/job/nightly-fortify-scan-ccd-1344/

### DIFF
--- a/src/uk/gov/hmcts/contino/AngularBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/AngularBuilder.groovy
@@ -16,6 +16,11 @@ class AngularBuilder extends AbstractBuilder {
   }
 
   @Override
+  def fortifyScan() {
+    builder.fortifyScan()
+  }
+
+  @Override
   def test() {
     builder.test()
   }
--- a/vars/fortifyScan.groovy
+++ b/vars/fortifyScan.groovy
@@ -5,18 +5,15 @@ import uk.gov.hmcts.contino.AppPipelineConfig
 
 def call(params) {
   PipelineCallbacksRunner pcr = params.pipelineCallbacksRunner
-  AppPipelineConfig config = params.appPipelineConfig
   Builder builder = params.builder
 
-  def environment = params.environment
   def product = params.product
+  def fortifyVaultName = params.fortifyVaultName
 
-  if (config.fortifyScan) {
-    stageWithAgent("Fortify Scan", product) {
-      withFortifySecrets(product, environment) {
-        pcr.callAround('fortifyscan') {
-          builder.fortifyScan()
-        }
+  stageWithAgent("Fortify Scan", product) {
+    withFortifySecrets(fortifyVaultName) {
+      pcr.callAround('fortify-scan') {
+        builder.fortifyScan()
       }
     }
   }
--- a/vars/fortifyScan.groovy
+++ b/vars/fortifyScan.groovy
@@ -1,0 +1,23 @@
+#!groovy
+import uk.gov.hmcts.contino.Builder
+import uk.gov.hmcts.contino.PipelineCallbacksRunner
+import uk.gov.hmcts.contino.AppPipelineConfig
+
+def call(params) {
+  PipelineCallbacksRunner pcr = params.pipelineCallbacksRunner
+  AppPipelineConfig config = params.appPipelineConfig
+  Builder builder = params.builder
+
+  def environment = params.environment
+  def product = params.product
+
+  if (config.fortifyScan) {
+    stageWithAgent("Fortify Scan", product) {
+      withFortifySecrets(product, environment) {
+        pcr.callAround('fortifyscan') {
+          builder.fortifyScan()
+        }
+      }
+    }
+  }
+}
--- a/vars/sectionNightlyTests.groovy
+++ b/vars/sectionNightlyTests.groovy
@@ -38,6 +38,14 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
       }
     }
 
+    fortifyScan(
+      pipelineCallbacksRunner: pcr,
+      appPipelineConfig: config,
+      builder: builder,
+      environment: environment.nonProdName,
+      product: product,
+    )
+
     if (config.crossBrowserTest) {
       stageWithAgent("Cross browser tests", product) {
         warnError('Failure in crossBrowserTest') {
@@ -104,6 +112,14 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
         }
       }
     }
+
+    highLevelDataSetup(
+      appPipelineConfig: config,
+      pipelineCallbacksRunner: pcr,
+      builder: builder,
+      environment: environment.nonProdName,
+      product: product,
+    )
 
     if (config.fullFunctionalTest) {
       stageWithAgent('Full functional tests', product) {
--- a/vars/sectionNightlyTests.groovy
+++ b/vars/sectionNightlyTests.groovy
@@ -38,13 +38,14 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
       }
     }
 
-    fortifyScan(
-      pipelineCallbacksRunner: pcr,
-      appPipelineConfig: config,
-      builder: builder,
-      environment: environment.nonProdName,
-      product: product,
-    )
+    if (config.fortifyScan) {
+      fortifyScan(
+        pipelineCallbacksRunner: pcr,
+        fortifyVaultName: config.fortifyVaultName ?: "${product}-${environment.nonProdName}",
+        builder: builder,
+        product: product,
+      )
+    }
 
     if (config.crossBrowserTest) {
       stageWithAgent("Cross browser tests", product) {
--- a/vars/withFortifySecrets.groovy
+++ b/vars/withFortifySecrets.groovy
@@ -1,10 +1,10 @@
-def call(String product, String environment, Closure block) {
+def call(String fortifyVaultName, Closure block) {
   def fortifySecrets = [
     [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-username', version: '', envVariable: 'FORTIFY_USER_NAME'],
     [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-password', version: '', envVariable: 'FORTIFY_PASSWORD'],
   ]
 
-  String theKeyVaultUrl = "https://${product}-${environment}.vault.azure.net/"
+  String theKeyVaultUrl = "https://${fortifyVaultName}.vault.azure.net"
 
   withAzureKeyvault(
     azureKeyVaultSecrets: fortifySecrets,
--- a/vars/withFortifySecrets.groovy
+++ b/vars/withFortifySecrets.groovy
@@ -1,0 +1,15 @@
+def call(String product, String environment, Closure block) {
+  def fortifySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-username', version: '', envVariable: 'FORTIFY_USER_NAME'],
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-password', version: '', envVariable: 'FORTIFY_PASSWORD'],
+  ]
+
+  String theKeyVaultUrl = "https://${product}-${environment}.vault.azure.net/"
+
+  withAzureKeyvault(
+    azureKeyVaultSecrets: fortifySecrets,
+    keyVaultURLOverride: theKeyVaultUrl
+  ) {
+    block.call()
+  }
+}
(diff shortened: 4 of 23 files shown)


vars/fortifyScan.groovy
```diff
@@ -0,0 +1,23 @@
...
+  AppPipelineConfig config = params.appPipelineConfig
+  Builder builder = params.builder
+
+  def environment = params.environment
+  def product = params.product
+
+  if (config.fortifyScan) {
+    stageWithAgent("Fortify Scan", product) {
+      withFortifySecrets(product, environment) {
+        pcr.callAround('fortifyscan') {
```
MSancaktutar, line 17: @timja  - Is the syntactic convention to use an all-lower case key without a space, in this line? eg `fortifyscan`

timja: no space, a dash can be used though

```suggestion
        pcr.callAround('fortify-scan') {
```

src/uk/gov/hmcts/contino/AngularBuilder.groovy
```diff
@@ -15,6 +15,11 @@ class AngularBuilder extends AbstractBuilder {
     builder.yarn("build:ssr")
   }
 
+  @Override
```
timja, line 18: the code in the PR looks fine, a couple of questions though:

1. Why create a java client rather than a bash script, it looks like some files need zipping and then a java jar invoked?
2. Why use `Runtime` to call a java jar on the command line, wouldn't it be less effort to add the jar to the classpath and call the code directly? rather than messing with process args, and string concatentation

MSancaktutar: Thank you @timja. Answers are as below:

1- There is and will be a significant deal of further logic that would be much nicer and better maintainable in java. (Simple html report generation, properties management, Jira integration with rest-apis, etc) Obviously, all those can still be implemented as a script, but much more convenient in a strongly typed, OOP way with enablement of huge deals of open source provisions for any later needs.

2- I had asked the same to Kaan, and he clarified that the code in the jar is written to check a particular jar version from which it is being executed, hence the work around. Otherwise, we'd actually call an object interface within the JVM. However, our prospect is that the original fortify client jar we are wrapping will later on be published in a more convenient way, along with more open support for code re-use, and then the hmcts' fortify client wrapper can be refactored seamlessly.

vars/withFortifySecrets.groovy
```diff
@@ -0,0 +1,15 @@
+def call(String product, String environment, Closure block) {
+  def fortifySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-username', version: '', envVariable: 'FORTIFY_USER_NAME'],
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-password', version: '', envVariable: 'FORTIFY_PASSWORD'],
+  ]
+
+  String theKeyVaultUrl = "https://${product}-${environment}.vault.azure.net/"
```
adusumillipraveen, line 7: Are these secrets specific to a team ? Why not just put it in the Jenkins vault?

kaanaktas: Yes they are. Each team will use their own Fortify secrets.

adusumillipraveen: Not sure if every team has a shared key vault with `${product}-${environment}` though.

kaanaktas: We request this in the document. So, they will create 2 secrets on Azure Key Vaults with their Fortify creds. @MSancaktutar

vars/fortifyScan.groovy
```diff
@@ -0,0 +1,23 @@
...
+  PipelineCallbacksRunner pcr = params.pipelineCallbacksRunner
+  AppPipelineConfig config = params.appPipelineConfig
+  Builder builder = params.builder
+
+  def environment = params.environment
+  def product = params.product
+
+  if (config.fortifyScan) {
+    stageWithAgent("Fortify Scan", product) {
+      withFortifySecrets(product, environment) {
```
njrich28, line 16: This is for static analysis of source code which is non-environment specific, so why pass the environment in here? It may lead to confusion.

From what I can see the only reason to pass it in is to look up a key vault using a hardcoded vault naming convention. Better to just pass the specific vault name in here.

vars/withFortifySecrets.groovy
njrich28: I echo what @adusumillipraveen says - we shouldn't assume or require that all products have a vault with this naming convention. Coupled with the comment above about passing around the environment for non-environment-specific tests it would be better to pass in the vault name and let users decide.

kaanaktas: I will have a look this.

vars/withFortifySecrets.groovy
```diff
@@ -0,0 +1,15 @@
+def call(String product, String fortifyVaultName, Closure block) {
+  def fortifySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-username', version: '', envVariable: 'FORTIFY_USER_NAME'],
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-password', version: '', envVariable: 'FORTIFY_PASSWORD'],
+  ]
+
+  String theKeyVaultUrl = fortifyVaultName ? "https://${fortifyVaultName}.vault.azure.net" : "https://${product}-aat.vault.azure.net/"
```
timja, line 7: needs to use the environment.nonProdName field, can't assume it will be called `aat`

kaanaktas: done

vars/withFortifySecrets.groovy
kaanaktas: This is done and has been tested in user-profile. Now each client is able to pass its own keyvaultname. 

https://build.platform.hmcts.net/view/CDM/job/HMCTS_Nightly_CDM/job/ccd-user-profile-api/job/nightly-fod-scan-ccd-1257/

MSancaktutar:
@njrich28, @adusumillipraveen , @timja  - could one of you please approve and merge the PR, as we need this merged as soon as possible really?

MSancaktutar (approved):
I think this is good to go after actioning the latest comments.

vars/sectionNightlyTests.groovy
```diff
@@ -38,6 +38,14 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
       }
     }
 
+    fortifyScan(
+      pipelineCallbacksRunner: pcr,
+      appPipelineConfig: config,
+      builder: builder,
+      environment: environment.nonProdName,
```
adusumillipraveen, line 45: Agree with Neil's previous comment, even i think we should get rid of `environment` from here even if it means we should make setting keyvault mandatory for everyone. This makes it look that his step is environment specific while it's not.

adusumillipraveen (commented):
Can you also add a readme for this, just a couple of lines with how to configure will do.  Recent example https://github.com/hmcts/cnp-jenkins-library#high-level-data-setup

vars/sectionNightlyTests.groovy
MSancaktutar: The implementation is parameterised to accept a different environment later on. Could that be treated a non-blocker, minor comment?

adusumillipraveen: sure, if others are happy with it.

timja: does this need a team specific key vault? or can it just use a central username and password?

MSancaktutar: It is indeed team specific.

njrich28: We should remove references to an `environment` - the check doesn't depend on _any_ environment. It's not a question of allowing teams to specify a different environment - including the `environment` as a parameter may make people think it's environment specific.

timja: You could pull up passing the vault name to here, which means you don't need to force everyone to set the vault name and can still rely on convention.

That may mean you need to pull up the `if enabled` check to around this function

vars/withFortifySecrets.groovy
```diff
@@ -0,0 +1,15 @@
+def call(String product, String environment, String fortifyVaultName, Closure block) {
```
timja, line 1: with https://github.com/hmcts/cnp-jenkins-library/pull/755/files#r665241989 this becomes
```suggestion
def call(String fortifyVaultName, Closure block) {
```

kaanaktas: done

vars/sectionNightlyTests.groovy
kaanaktas: Environment variable has been removed and we now completely rely on the vaultValue which is passed through enableFortifyScan(). Can you please have a look? By the way, some tests are failing in the pipeline, but not in my local.

https://sandbox-build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_Sandbox_Pipeline_Test%2Fcnp-jenkins-library/detail/PR-755/17/tests

timja: is there a reason you didn't pass the vault name down from here allowing a convention based approach like suggested above?

MSancaktutar: @timja, Kaan and I have got a little confused amongst various comments. Could you please suggest a commit or just push one to the branch? Could you please also resolve other comments you were satisfied with the actions/responses for?

kaanaktas: @timja I have tried to follow the same login which has been applied to enableDbMigration('ccd'). But I don't understand if we don't force people to pass it, what we would put for the vault name. 
I guess you want me to make below change. So, how should we handle vault name if users don't pass it

    if (config.fortifyScan) {
      fortifyScan(
        pipelineCallbacksRunner: pcr,
        fortifyVaultName: config.fortifyVaultName,
        builder: builder,
        product: product,
      )
    }

timja: something like:

```
if (config.fortifyScan) {
  fortifyScan(
    pipelineCallbacksRunner: pcr,
    fortifyVaultName: config.fortifyVaultName ?: "https://${var.product}-${environment.nonProdName}.vault.azure.net/",
    builder: builder,
    product: product,
  )
}
```

kaanaktas: We can do that. What I understand from @njrich28 comment [here](https://github.com/hmcts/cnp-jenkins-library/pull/755#discussion_r665229566), we shouldn't use environment name anywhere in the vault url, as this make people confused. Now, I am making the above change.

vars/sectionNightlyTests.groovy
```diff
@@ -38,6 +38,15 @@ def call(PipelineCallbacksRunner pcr, AppPipelineConfig config, PipelineType pip
       }
     }
 
+    if (config.fortifyScan) {
+      fortifyScan(
+        pipelineCallbacksRunner: pcr,
+        fortifyVaultUrl: config.fortifyVaultUrl ?: "https://${product}-${environment.nonProdName}.vault.azure.net/",
```
timja, line 44: please use vault name not url,

We have no need for people to use something other than vault.azure.net domain and it makes the pipeline config a lot more verbose and is also different to db migration

https://github.com/hmcts/cnp-jenkins-library/blob/ccd-1257/src/uk/gov/hmcts/contino/AppPipelineConfig.groovy#L8

```suggestion
        fortifyVaultName: config.fortifyVaultName ?: "${product}-${environment.nonProdName}",
```

vars/withFortifySecrets.groovy
```diff
@@ -0,0 +1,13 @@
+def call(String fortifyVaultUrl, Closure block) {
+  def fortifySecrets = [
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-username', version: '', envVariable: 'FORTIFY_USER_NAME'],
+    [$class: 'AzureKeyVaultSecret', secretType: 'Secret', name: 'fortify-on-demand-password', version: '', envVariable: 'FORTIFY_PASSWORD'],
+  ]
+
+  withAzureKeyvault(
+    azureKeyVaultSecrets: fortifySecrets,
+    keyVaultURLOverride: fortifyVaultUrl
```
timja, line 9:
```suggestion
    keyVaultURLOverride: "https://${fortifyVaultName}.vault.azure.net/"
```

timja (changes requested):
please change to vault name, have added a couple of suggestions to clarify but not all places

vars/sectionNightlyTests.groovy
kaanaktas: done

timja (approved):

MSancaktutar (approved):
Thanks for the review comments. I think this is now good to go.

njrich28 (approved):

MSancaktutar:
Thanks a lot @kaanaktas for the beautiful piece of work.
Many thanks for the review comments and contributions @timja , @njrich28 , @adusumillipraveen .

PR #758, opened 2021-07-02
RWA-608 Create test service via test DMN files

mkayad:
[RWA-608](https://tools.hmcts.net/jira/browse/RWA-608)


louisehuyton:
Hi @mkayad, I notice that the checks are failing. Platform Operations cannot merge this ticket until these checks are passing.

PR #759, opened 2021-07-05
Added support to disable high level data setup

nitinprabhu:
* Currently if we enable high level data setup it enables it for AAT and PROD and expects key vaults to be present in prod else it fails for reference https://build.platform.hmcts.net/view/NFDIV/job/HMCTS_NFDIV/job/nfdiv-case-api/job/master/503/console
* For services such which are not in prod and only in AAT this fails the build hence adding support to disable the high level data setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,21 @@ Branch | HighDataSetup Stage
 `demo` | `demo`
 `ithc` | `ithc`
 
+To disable high level data setup for an environment use
+```
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "java"
+def product = "rhubarb"
+def component = "recipe-backend"
+
+withPipeline(type, product, component) {
+  disableHighLevelDataSetup()
+}
+
+```
 #### Extending the opinionated pipeline
 
 It is not possible to remove stages from the pipeline but it is possible to _add_ extra steps to the existing stages.
--- a/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
+++ b/src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
@@ -101,4 +101,8 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
   void enableHighLevelDataSetup() {
     config.highLevelDataSetup = true
   }
+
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -189,6 +189,21 @@ Branch | HighDataSetup Stage
...
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "java"
+def product = "rhubarb"
+def component = "recipe-backend"
+
+withPipeline(type, product, component) {
+  disableHighLevelDataSetup()
```
timja, line 203: how does this disable for only an environment? it's disabled by default anyway?

nitinprabhu: It's disabled by default but if we enable it then it runs on AAT and PROD on master. Currently we are only in AAT so we don't want this to be executed for prod.
So I am thinking of let it execute on AAT and then before prod set is run will disable it 

will below work ?

```
    after('aksdeploy:aat') {
      disableHighLevelDataSetup()
    }
```

timja: if you're only in AAT then won't the pipeline not run prod?

i.e. why not run it in prod?

nitinprabhu: We don't have resources such as key vault created in vault so build pipeline fails as the high level data setup tasks expects importer secrets in prod vault
```
14:27:16  Caused: org.jenkinsci.plugins.azurekeyvaultplugin.AzureKeyVaultException: Failed to retrieve secret definition-importer-username from vault https://nfdiv-prod.vault.azure.net/, error message: Max retries 3 times exceeded. Error Details: nfdiv-prod.vault.azure.net
```

nitinprabhu: https://build.platform.hmcts.net/view/NFDIV/job/HMCTS_NFDIV/job/nfdiv-case-api/job/master/503/console

timja: the fix should be to re-structure the pipeline so this is correctly wrapped around the approved for environment, if you look at the pipeline you'll see build infra not running in prod so just make sure that the same logic is applied to this.

you could also just deploy them too

nitinprabhu: ok cool will close this PR then thanks for your inputs

PR #760, opened 2021-07-05
DTSPO-2879 - SDS Jenkins setup

endakelly:
Adding environment variable for sbox which is required in sds jenkins setup


timja (changes requested):
you should be setting `SANDBOX_ENVIRONMENT_NAME` in our flux config, adding this won't do anything

PR #772, opened 2021-08-09
DTSPO-2880 - adding ithc and perftest subs for SDS

endakelly:
Existing CFT branches should continue to default to 'qa' as QA_SUBSCRIPTION_NAME does not seem to be used anywhere


msl8r (approved):
LGTM

PR #777, opened 2021-09-13
add optional param and clear dependencies

chris-pateman:
This change should clear the cache of gradle builds, so there is no missing dependencies.

https://tools.hmcts.net/jira/browse/DTSPO-4836

### DIFF
--- a/.bash_history
+++ b/.bash_history
@@ -1,8 +1,0 @@
-ls
-gradlew build
-./gradlew build
-./gradlew build
-cd gradlew
-ls
-gradlew
-exit
--- a/.bash_history
+++ b/.bash_history
@@ -1,0 +1,8 @@
+ls
+gradlew build
+./gradlew build
+./gradlew build
+cd gradlew
+ls
+gradlew
+exit
--- a/src/uk/gov/hmcts/contino/GradleBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/GradleBuilder.groovy
@@ -15,7 +15,7 @@ class GradleBuilder extends AbstractBuilder {
 
   def build() {
     addVersionInfo()
-    gradle("assemble")
+    gradle("assemble", additionalOptions = "--refresh-dependencies")
   }
 
   def fortifyScan() {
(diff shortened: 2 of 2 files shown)


.bash_history
```diff
@@ -0,0 +1,8 @@
+ls
```
timja, line 1: I don't think you meant to commit this?

src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -15,7 +15,7 @@ class GradleBuilder extends AbstractBuilder {
 
   def build() {
     addVersionInfo()
-    gradle("assemble")
+    gradle("assemble", additionalOptions = "--refresh-dependencies")
```
timja, line 18: this seems to say that it ignores downloaded dependencies, what's the reason for this?
https://stackoverflow.com/a/31832995/4951015

If it's a one-off build agent issue the fix should be just to delete the build agent.

This will likely slow builds down a bit but probably not too badly as we are caching dependencies at artifactory

.bash_history
chris-pateman: no, thanks for highlighting

src/uk/gov/hmcts/contino/GradleBuilder.groovy
chris-pateman: @timja can't say I am 100% with Java, so I am taking advice from the developer who said it is needed due to dependencies not downloading correctly into the container: https://tools.hmcts.net/jira/browse/DTSPO-4836

how do we remove a build agent instead?

timja: just ask on platops-help and we can delete the agent.

src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -174,12 +174,12 @@ EOF
...
 
 
-  def gradle(String task, String prepend = "") {
+  def gradle(String task, String prepend = "", String additionalOptions = "") {
     if (prepend && !prepend.endsWith(' ')) {
       prepend += ' '
     }
     addInitScript()
-    steps.sh("${prepend}./gradlew --no-daemon --init-script init.gradle ${task}")
+    steps.sh("${prepend}./gradlew --no-daemon --init-script ${additionalOptions} init.gradle ${task}")
```
timja, line 182: this also isn't the right place anyway, it needs to go after the `init.gradle`

chris-pateman:
Requesting Agent Deleted instead of Code Change

PR #785, opened 2021-11-17
add optional component parameter for infrastructure

chris-pateman:
Add the optional parameter component to infrastructure pipelines.
When having more then one pipeline that only deploys infrastructure they clash as they use the same state file using the product name. This will give the option to pass in the component name, which is used in the state file location name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,9 +278,9 @@ withInfraPipeline(product) {
 You have the ability to pass extra parameters to the `withInfraPipeline`.
 
 These parameters include:
-| parameter name | description | example |
+| parameter name | description
 | --- | --- | --- |
-| component | Extra detail to the name of the repository | "extra-detail" |
+| component | https://hmcts.github.io/glossary/#component |
 
 Example `Jenkinsfile` to use the opinionated infrastructure pipeline:
 ```groovy
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ Example `Jenkinsfile` to use the opinionated infrastructure pipeline:
 
 @Library("Infrastructure") _
 
-def product = "rhubarb" 
+def product = "rhubarb"
 
 withInfraPipeline(product) {
 
--- a/README.md
+++ b/README.md
@@ -276,6 +276,33 @@ withInfraPipeline(product, component) {
 }
 ```
 
+#### Optional parameters for the opinionated infratructure pipeline
+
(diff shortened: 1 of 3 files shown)


vars/withInfraPipeline.groovy
```diff
@@ -7,7 +7,7 @@ import uk.gov.hmcts.contino.Subscription
 import uk.gov.hmcts.contino.Environment
 import uk.gov.hmcts.pipeline.TeamConfig
 
-def call(String product, Closure body) {
+def call(String product, String component = "", Closure body) {
```
timja, line 10: use null not empty string
```suggestion
def call(String product, String component = null, Closure body) {
```

README.md
```diff
@@ -266,7 +266,10 @@ Example `Jenkinsfile` to use the opinionated infrastructure pipeline:
 
 def product = "rhubarb"
 
-withInfraPipeline(product) {
+//Optional
+def component = "extra-detail"
```
timja, line 270: this should be a separate example, people tend to copy paste and given it's not our default approach then it shouldn't be included in this example

chris-pateman: I have added a new section. please review

vars/withInfraPipeline.groovy
chris-pateman: updated

README.md
```diff
@@ -264,7 +264,7 @@ Example `Jenkinsfile` to use the opinionated infrastructure pipeline:
 
 @Library("Infrastructure") _
 
-def product = "rhubarb"
+def product = "rhubarb"
```
timja, line 267:
```suggestion
def product = "rhubarb"
```

README.md
```diff
@@ -273,6 +273,33 @@ withInfraPipeline(product) {
...
 ```
 
+#### Optional parameters for the opinionated infratructure pipeline
+
+You have the ability to pass extra parameters to the `withInfraPipeline`.
+
+These parameters include:
+| parameter name | description | example |
+| --- | --- | --- |
+| component | Extra detail to the name of the repository | "extra-detail" |
```
timja, line 283: not the best name really for the description, wondering if should just link to the glossary?
https://hmcts.github.io/glossary/#component

timja (dismissed):

timja (approved):

PR #814, opened 2022-02-23
SIDM-7016 Add saat variable declaration, Environment test.

dattipoeHMCTS:
Notes:
*
*
*

PR #815, opened 2022-02-24
Sidm 7016

verityit:
Missed def saat

Notes:
*
*
*


timja:
Not needed as of https://github.com/hmcts/cnp-jenkins-library/pull/814

PR #817, opened 2022-03-02
DTSPO-6900.  Using old ZAPproxy image

mokainos:
https://github.com/zaproxy/zaproxy/issues/7111


timja (approved):
this is okay for now to unblock if you've tested it in a pipeline but we need a solution so we aren't stuck here forever

thc202:
There's an update in the referenced issue.

timja (changes requested):
let's hold off on this

PR #856, opened 2022-06-20
Restrict renovate packages

timja:


renovate.json
```diff
@@ -1,5 +1,15 @@
 {
   "extends": [
     "config:base"
+  ],
+  "packageRules": [
+    {
+      "matchPackageNames": ["org.codehaus.groovy:groovy-dateutil", "org.codehaus.groovy:groovy-all"],
+      "allowedVersions": "<3"
```
timja, line 8: Both spock and JenkinsPipelineUnit depend on groovy.

JenkinsPipelineUnit seems to track the Jenkins groovy version, forcing it to be upgraded seems risky.
so lets not force it to use above 3 till JenkinsPipelineUnit uses 3.x

renovate.json
```diff
@@ -1,5 +1,15 @@
...
+    },
+    {
+      "matchPackageNames": ["org.spockframework:spock-core"],
+      "allowedVersions": "/^[0-9]+\\.[0-9]+(\\.[0-9]+)?-groovy-2\\.5$/"
```
timja, line 12: spock provides per version suffix groovy versions, it would be easier if these were different artifacts as renovate tries to update to groovy 3.x with this 😢 

hopefully this works, I tested via regex tester and looked at the docs 
https://docs.renovatebot.com/configuration-options/#allowedversions

adusumillipraveen (approved):

PR #860, opened 2022-06-22
Cache results of getLabels() to reduce API calls

BWeeks-JusticeGovUK:
Cache labels to reduce API calls.

refreshLabelCache() makes an API call to update the cache.  Called automatically if cache is deemed invalid.

clearLabelCache() empties the cache.  The next getLabelsXXX() or addLabelsXXX() call will trigger a refresh.  It should not be necessary to call this method.

Private method getLabelsFromCache() either returns the cached list, or if empty, calls refreshLabelCache().

getLabels() calls getLabelsFromCache() instead of making a direct API call.

getLabelsbyPattern(branch_name, key) should be utilised when an ArrayList of multiple labels is required/expected.  Returns an ArrayList of results.

checkForLabel(branch_name, key) should be utilised when a single label is required.  Returns whether the label is present (boolean)

### DIFF
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -150,15 +150,22 @@ class GithubAPI {
   }
 
   /**
+   * Check for Pull Request
+   */
+  private isPR(String branch_name) {
+    def isPr = new ProjectBranch(branch_name).isPR()
+    this.steps.echo "Is ${branch_name} a PR?: ${isPr}"
+    return isPr
+  }
+
+  /**
    * Check Pull Request for label by a pattern in name.
    */
   def getLabelsbyPattern(String branch_name, String key) {
     this.steps.echo "Getting Labels for Branch: ${branch_name} by Pattern: ${key}"
-    def isPR = new ProjectBranch(branch_name).isPR()
-    this.steps.echo "Is this a PR?: ${isPR}"
-    if (isPR) {
+    if (isPR(branch_name)) {
       this.steps.echo "PR Confirmed.  Calling getLabels()."
-      def foundLabels = getLabels().findAll{it.contains(key)}
+      def foundLabels = getLabels(branch_name).findAll{it.contains(key)}
       this.steps.echo "Returning Labels: ${foundLabels}"
       return foundLabels
     } else {
@@ -180,9 +187,15 @@ class GithubAPI {
   /**
    * Get all labels from an issue or pull request
    */
-  def getLabels() {
-    this.steps.echo "Getting All Labels.  Calling getLabelsFromCache()."
-    return this.getLabelsFromCache()
+  def getLabels(String branch_name) {
+    this.steps.echo "Getting All Labels for ${branch_name}."
+    if (isPR(branch_name)) {
+      this.steps.echo "PR Confirmed.  Calling getLabelsFromCache()."
+      return this.getLabelsFromCache()
+    } else {
+      this.steps.echo "Negative PR.  Returning Empty List."
+      return []
+    }
   }
 
   def currentProject() {
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -195,15 +195,9 @@ class GithubAPI {
    */
   def getLabelsbyPattern(String branch_name, String key) {
     this.steps.echo "Getting Labels for Branch: ${branch_name} by Pattern: ${key}"
-    if (isPR(branch_name)) {
-      this.steps.echo "PR Confirmed.  Calling getLabels()."
-      def foundLabels = getLabels(branch_name).findAll{it.contains(key)}
-      this.steps.echo "Returning Labels: ${foundLabels}"
-      return foundLabels
-    } else {
-      this.steps.echo "Negative PR.  Returning Empty List."
-      return []
-    }
+    def foundLabels = getLabels(branch_name).findAll{it.contains(key)}
+    this.steps.echo "Returning Labels: ${foundLabels}"
+    return foundLabels
   }
 
   /**
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -62,6 +62,7 @@ class GithubAPI {
       if (this.cachedLabelList.isValid) {
         this.steps.echo "Cache is Valid.  Adding new labels to cache."
         this.cachedLabelList.cache.addAll(labels)
+        this.cachedLabelList.cache.unique()
         this.steps.echo "Updated Cache Contents: ${this.cachedLabelList.cache}"
       } else {
         this.steps.echo "Cache is Invalid.  Refreshing Cache."
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -85,6 +85,20 @@ class GithubAPI {
   }
 
   /**
+   * Return whether the cache is valid
+   */
+  def isCacheValid() {
+    return this.cachedLabelList.isValid
+  }
+
+  /**
+   * Return whether the chache is empty
+   */
+  def isCacheEmpty() {
+    return this.cachedLabelList.cache.isEmpty()
+  }
+
+  /**
    * Refreshes this.cachedLabelList
    */
   def refreshLabelCache() {
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -99,6 +99,13 @@ class GithubAPI {
   }
 
   /**
+   * Return cache contents as list
+   */
+  def getCache() {
+    return this.cachedLabelList.cache
+  }
+
+  /**
    * Refreshes this.cachedLabelList
    */
   def refreshLabelCache() {
--- a/src/uk/gov/hmcts/contino/GithubAPI.groovy
+++ b/src/uk/gov/hmcts/contino/GithubAPI.groovy
@@ -38,22 +38,22 @@ class GithubAPI {
   /**
    * Return whether the cache is valid
    */
-  def isCacheValid() {
-    return this.cachedLabelList.isValid
+  def static isCacheValid() {
+    return cachedLabelList.isValid
   }
 
   /**
-   * Return whether the chache is empty
+   * Return whether the cache is empty
    */
-  def isCacheEmpty() {
-    return this.cachedLabelList.cache.isEmpty()
+  def static isCacheEmpty() {
+    return cachedLabelList.cache.isEmpty()
   }
 
   /**
    * Return cache contents as list
    */
-  def getCache() {
-    return this.cachedLabelList.cache
+  def static getCache() {
+    return cachedLabelList.cache
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,6 +13,8 @@ class GithubAPI {
     this.steps = steps
   }
 
+  private cachedLabelList = []
```
timja, line 16: should be static

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -49,16 +52,30 @@ class GithubAPI {
...
+
 /**
  * Check Pull Request for label by a pattern in name.
  */
   def getLabelsbyPattern(String branch_name, String key) {
     if (new ProjectBranch(branch_name).isPR() == true) {
-      def project = currentProject()
-      def pullRequestNumber = currentPullRequestNumber()
-      return getLabels(project, pullRequestNumber).findAll{it.contains(key)}
+      return this.getLabelsFromCache().findAll{it.contains(key)}
```
timja, line 78: is there a reason you didn't change `getLabels` internally? Given it's part of the public API it would be good if it was transitioned to use the cache as well for safety I think

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -49,16 +52,30 @@ class GithubAPI {
       requestBody: "${body}",
       consoleLogResponseBody: true,
       validResponseCodes: '200')
+
+    this.cachedLabelList = []
```
timja, line 56: needs testing, not 100% sure `this` will work in a static context, may work though, https://github.com/hmcts/cnp-jenkins-library/blob/master/src/uk/gov/hmcts/pipeline/TeamConfig.groovy#L41 refers to it with no qualifier

src/uk/gov/hmcts/contino/GithubAPI.groovy
timja: there shouldn't be any need to update the cache, and if we need to we can add it I think.

BWeeks-JusticeGovUK: getLabels() now utilises the cache.  Added refreshLabelCache() which can be called to update the cache if required.

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: Looking at this further, I'm not at all certain that cachedLabelList should be static.  My understanding is that a static property exists once, rather than for each instance of a class - which would absolutely break this given that each instance will be returning it's own list of labels (from different Repo's) to store in the cache.

src/uk/gov/hmcts/contino/GithubAPI.groovy
timja, line 16: static is gone?

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -49,16 +52,47 @@ class GithubAPI {
...
+    def json_response = new JsonSlurper().parseText(response.content)
+    this.cachedLabelList = json_response.collect( { label -> label['name'] } )
+    return this.cachedLabelList
+  }
+
+/**
+ * Check this.cachedLabelList, if empty, call getLabels() to repopulate.
+*/
+  private getLabelsFromCache() {
+    if (this.cachedLabelList.size() == 0) {
```
timja, line 82:
```suggestion
    if (this.cachedLabelList.isEmpty()) {
```

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -49,16 +52,47 @@ class GithubAPI {
...
+      return refreshLabelCache()
```
timja, line 83: won't we still end up calling github multiple times if there's no labels?
shouldn't we use use null and init it to empty list or have a placeholder label that shows we haven't tried loading yet?

BWeeks-JusticeGovUK: Yeah good point - should be resolved with the latest commit.

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: Based on my understanding, I don't believe that we want it static.  Making it so will decouple the cache from each instance of the github api class, which will break if multiple instances are created for different repos.

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,6 +13,11 @@ class GithubAPI {
     this.steps = steps
   }
 
+  private cachedLabelList = [
```
timja, line 16: shouldn't this be static so it persists across the whole pipeline run? in case someone is creating a new instance at a different point?

BWeeks-JusticeGovUK: My understanding is that a static property is decoupled from the instance.  Therefore if we make it static, there will be one cache for all instances of the githubapi class within that pipeline, which we don't want (because each instance could point to a different repo).  Multiple instances of the API should not be created for the same repo within a given pipeline - the original api object should be reused instead.

timja: yes but there's no way for someone to get access to it outside this pipeline step we use so they would have to create their own instance in their Jenkinsfile. not sure why you would check another repo other than your own

BWeeks-JusticeGovUK: Ah I see - being unable to pass the original API object does make the case for static more compelling.  I'm not 100% sure if it will work as intended.  We should test with and without.

(WRE other repos, some PR's will stand up multiple pods for different applications.  I was thinking that if those additional pods are based on PR's it might be useful/necessary to check the labels on those also - probably a fringe use case though.)

timja: we do the static approach elsewhere, from my understanding static is for a pipeline run not Jenkins itself so it should be fine, but worth testing with 2 different repositories

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -49,37 +55,29 @@ class GithubAPI {
...
-   *
-   * @param project
-   *   The project repo name, including the org e.g. 'hmcts/my-frontend-app'
-   * @param issueNumber
-   *   The issue or PR number
+   * Clears this.cachedLabelList
    */
-  def getLabels(project, issueNumber) {
+  private void clearLabelCache() {
+    echo "Clearing Label Cache."
```
timja, line 66: this won't work it needs to be prefixed by `steps`
```suggestion
    steps.echo "Clearing Label Cache."
```

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,22 +13,94 @@ class GithubAPI {
...
-   * Add labels to the current pull request.  MUST be run with an onPR() closure.
-   *
-   * @param labels
-   *   A List of labels
+   * Check for Pull Request
    */
-  def addLabelsToCurrentPR(labels) {
+  private isPR(String branch_name) {
+    def isPr = new ProjectBranch(branch_name).isPR()
+    this.steps.echo "Is ${branch_name} a PR?: ${isPr}"
```
timja, line 34: is this still needed?

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,22 +13,94 @@ class GithubAPI {
...
+   * Return cache contents as list
+   */
+  def static getCache() {
+    return cachedLabelList.cache
+  }
+
+  /**
+   * Clears this.cachedLabelList
+   */
+  void clearLabelCache() {
```
timja, line 62: is this still needed? is it a future like method? I can't see it called outside of tests

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,22 +13,94 @@ class GithubAPI {
...
+    this.steps.echo "Clearing Label Cache."
+    cachedLabelList.cache = []
+    cachedLabelList.isValid = false
+    this.steps.echo "Cleared Cache Contents: ${getCache()}"
+    this.steps.echo "Cleared Cache Valid?: ${isCacheValid()}"
```
timja, line 67: this doesn't look needed at all as you just set it a couple of lines up

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -13,22 +13,94 @@ class GithubAPI {
...
+
+    this.steps.echo "Response Status Code: ${response.status}"
+
+    if (response.status == 200) {
+      this.steps.echo "Response Ok."
+      def json_response = new JsonSlurper().parseText(response.content)
+      cachedLabelList.cache = json_response.collect({ label -> label['name'] })
+      cachedLabelList.isValid = true
+      this.steps.echo "Cache Contents: ${getCache()}"
+      this.steps.echo "Cache Valid?: ${isCacheValid()}"
```
timja, line 93: do we need both these?

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -38,65 +110,113 @@ class GithubAPI {
    *   A List of labels
    */
   def addLabels(project, issueNumber, labels) {
-
+    this.steps.echo "Adding the following labels: ${labels}"
     def body = JsonOutput.toJson(labels)
-
+    this.steps.echo "Request Body: ${body}"
```
timja, line 115: do we need to log this still?

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -38,65 +110,113 @@ class GithubAPI {
...
-    if (new ProjectBranch(branch_name).isPR() == true) {
-      def project = currentProject()
-      def pullRequestNumber = currentPullRequestNumber()
-      return getLabels(project, pullRequestNumber).findAll{it.contains(key)}
+    this.steps.echo "Response Status Code: ${response.status}"
+
+    if (response.status == 200) {
+      this.steps.echo "Response Ok."
+      if (isCacheValid()) {
+        this.steps.echo "Cache is Valid.  Adding new labels to cache."
```
timja, line 130: do we need this?

src/uk/gov/hmcts/contino/GithubAPI.groovy
```diff
@@ -38,65 +110,113 @@ class GithubAPI {
...
+        cachedLabelList.cache.addAll(labels)
+        cachedLabelList.cache.unique()
+        this.steps.echo "Updated Cache Contents: ${getCache()}"
+      } else {
+        this.steps.echo "Cache is Invalid.  Refreshing Cache."
```
timja, line 135: also do we need this?
```suggestion
        this.steps.echo "Cache is Invalid. Refreshing Cache."
```

timja (commented):
looks good although big is failing, few questions around logging

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: No - the isPR() method has now been removed.

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: It's essential to clear up the cache between tests.  May as well leave it in place for future use if required - we have an exposed method for refreshing the cache, and I think having one to clear it makes sense.

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: No - removed.

src/uk/gov/hmcts/contino/GithubAPI.groovy
BWeeks-JusticeGovUK: Removed the additional logging (not required outside of initial testing)

PR #870, opened 2022-07-13
Dtspo 9278 jenkins library update

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-9278
Update jenkins library to set global disableTraefikTls flag options based on what's passed through flux. Eg. for sds - https://github.com/hmcts/sds-flux-config/pull/2042

Tested this library against multiple places, see inside DTSPO ticket for pipelines

Notes:
*
*
*

### DIFF
--- a/vars/helmInstall.groovy
+++ b/vars/helmInstall.groovy
@@ -112,13 +112,6 @@ def call(DockerImage dockerImage, Map params) {
       "--namespace ${namespace}"
     ]
 
-    if (this.env.DISABLE_TRAEFIK_TLS == "true") {
-      //Add global disableTraefikTls flag as true if DISABLE_TRAEFIK_TLS is set to true in jenkins
-        options.add("--set global.disableTraefikTls=true")
-    } else {
-        options.add("--set global.disableTraefikTls=false")
-    }
-
     if (!config.serviceApp) {
       //Forcing Jobs deployed through Jenkins to be Job to avoid cronJobs being run forever.
         options.add("--set global.job.kind=Job")


adusumillipraveen (dismissed):

vars/helmInstall.groovy
```diff
@@ -111,6 +111,13 @@ def call(DockerImage dockerImage, Map params) {
       "--namespace ${namespace}"
     ]
 
+    if (this.env.DISABLE_TRAEFIK_TLS == "true") {
+      //Add global disableTraefikTls flag as true if DISABLE_TRAEFIK_TLS is set to true in jenkins
+        options.add("--set global.disableTraefikTls=true")
+    } else {
+        options.add("--set global.disableTraefikTls=false")
+    }
+
```
timja, line 120: this is not correct, why would you set an env value that would break all installs?
```suggestion
```

vars/helmInstall.groovy
```diff
@@ -111,6 +111,13 @@ def call(DockerImage dockerImage, Map params) {
       "--namespace ${namespace}"
```
timja, line 111:
```suggestion
      "--set global.disableTraefikTls=false",
      "--namespace ${namespace}"
```

timja (changes requested):
I would think about renaming the flag in a new version as the disable is a bit confusing given it's off my default.

`enableTraefikTls=true` would be clearer or `tls=true`

or
`ingress.tls.enabled=true`

adusumillipraveen (approved):

PR #883, opened 2022-08-03
DTSPO-6455 - Add support for yarn v3

JayDoubleu:
This adds support for yarn v3 while maintaining compatibility for yarn v1.

[corepack](https://github.com/nodejs/corepack#-corepack ) is used to install yarn v3 on-demand

For example upgrade PR see https://github.com/hmcts/sds-toffee-frontend/pull/52
Make sure you're using the node version >14 which includes corepack.

We recommend to use [zero-installs](https://yarnpkg.com/features/zero-installs) this will mean that you don't need to run `yarn install` when you swap between branches and the network will not need to be contacted as all files are checked into the repo (unless your app on-demand downloads something like puppeteer, this will still work with zero-installs though).

Yarn v1 entered maintenance mode in January 2020 and has received no new features since then.

It is recommended to upgrade so that you stay on a modern tool chain and get the performance benefits that come with it.

The express.js templates will be updated next sprint so that all new apps use yarn v3

### DIFF
--- a/src/uk/gov/hmcts/contino/YarnBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/YarnBuilder.groovy
@@ -242,7 +242,10 @@ EOF
     yarn("test:can-i-deploy:consumer")
   }
 
-  private runYarn(task){
+  private runYarn(String task, String prepend = ""){
+    if (prepend && !prepend.endsWith(' ')) {
+      prepend += ' '
+    }
     boolean yarnV2OrNewer = isYarnV2OrNewer()
     if (steps.fileExists(NVMRC)) {
       steps.sh """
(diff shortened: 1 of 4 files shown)


src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -248,9 +248,20 @@ EOF
...
+    def status = steps.sh label:"Determine if is YarnV1", script: '''
+                ! grep \"packageManager\": package.json
+          ''', returnStatus: true
+    return status
+  }
+
   def yarn(task) {
     if (!steps.fileExists(INSTALL_CHECK_FILE) && !runYarnQuiet("check")) {
-      runYarn("--mutex network install --frozen-lockfile")
+      if (!isYarnV1()) {
```
timja, line 260: A personal preference but if the positive case was first I find it easier to read, it could be something like:

```suggestion
      if (isYarnV2OrNewer()) {
```
or
flip the if statements around

there's some discussion around here:
https://softwareengineering.stackexchange.com/a/350551

which generally agree that positive is easier although not to do it at the expense of making it un-readable

timja (dismissed):

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -248,12 +248,25 @@ EOF
     return status == 0  // only a 0 return status is success
   }
 
+  private yarnV2OrNewer() {
```
timja, line 251:
```suggestion
  private isYarnV2OrNewer() {
```

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -248,12 +248,25 @@ EOF
...
-    runYarn(task)
+      boolean yarnV2OrNewer = isYarnV2OrNewer()
+          if (yarnV2OrNewer && !steps.fileExists(INSTALL_CHECK_FILE)) {
+              steps.sh """
+                  corepack enable
+                  touch ${INSTALL_CHECK_FILE}
+              """
+          } else if (!yarnV2OrNewer && !steps.fileExists(INSTALL_CHECK_FILE) && !runYarnQuiet("check")) {
+              runYarn("--mutex network install --frozen-lockfile")
+                  steps.sh("touch ${INSTALL_CHECK_FILE}")
```
timja, line 267: formatting is off here

src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -111,31 +111,47 @@ class YarnBuilder extends AbstractBuilder {
...
+            fi
+            chmod +x yarn-audit-with-suppressions.sh
+            ./yarn-audit-with-suppressions.sh
+          """
+      } finally {
+            if (yarnV2OrNewer) {
+              steps.sh """
+              cat yarn-audit-result | jq -c '. | {type: "auditSummary", data: .metadata}' > yarn-audit-issues-result-summary
+              cat yarn-audit-result | jq -cr '.advisories| to_entries[] | {"type": "auditAdvisory", "data": { "advisory": .value }}' >> yarn-audit-issues-advisories
+              cat  yarn-audit-issues-result-summary  yarn-audit-issues-advisories > yarn-audit-issues-result
```
timja, line 143:
```suggestion
              cat yarn-audit-issues-result-summary yarn-audit-issues-advisories > yarn-audit-issues-result
```

timja (approved):

PR #908, opened 2022-10-05
Updating to use zap2docker-stable instead of zap2docker-weekly

willwatters:

Notes:
*
*
*


timja (commented):
could we pin it to an image version as well? won't the stable version just break at some point?

msl8r (dismissed):

willwatters:
> 



> could we pin it to an image version as well? won't the stable version just break at some point?

Yes - I have updated as follows: owasp/zap2docker-stable:2.11.1

timja (approved):

cakeben (approved):

PR #923, opened 2022-11-08
Disable docker steps if docker file not present

adusumillipraveen:
Notes:
* withPipeline is currently used for repos like ccd-definition, camunda-definition and they add a dummy docker file which goes through entire lifecycle in the pipeline for no reason.

### DIFF
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -36,7 +36,7 @@ def call(params) {
       echo("Checking if we should skip image build, tag: ${projectBranch.imageTag()}, git commit: ${env.GIT_COMMIT}, timestamp: ${env.LAST_COMMIT_TIMESTAMP}, hasTag: ${hasTag}, hasOverride: ${envOverrideForSkip}, result: ${!noSkipImgBuild}")
     }
   }
-
+  env.DOCKER_IMAGE_EXISTS= fileExists('Dockerfile')
   onPathToLive {
     stageWithAgent("Build", product) {
       onPR {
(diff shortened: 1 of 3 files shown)


vars/sectionBuildAndTest.groovy
```diff
@@ -89,35 +84,37 @@ def call(params) {
...
-                      echo -e '\n' >> .dockerignore
-                      cat .dockerignore_build >> .dockerignore
-                    """
-            }
-            def buildArgs = projectBranch.isPR() ? " --build-arg DEV_MODE=true" : ""
-            if (fileExists(acbTemplateFilePath)) {
-              acr.runWithTemplate(acbTemplateFilePath, dockerImage)
-            } else {
-              acr.build(dockerImage, buildArgs)
+    if (fileExists('Dockerfile')) {
```
timja, line 87: can you extract to a variable so we aren't checking this multiple times in this section?

vars/sectionDeployToAKS.groovy
```diff
@@ -41,8 +41,10 @@ def call(params) {
     imageRegistry = env.TEAM_CONTAINER_REGISTRY ?: env.REGISTRY_NAME
     acr = new Acr(this, subscription, imageRegistry, env.REGISTRY_RESOURCE_GROUP, env.REGISTRY_SUBSCRIPTION)
     dockerImage = new DockerImage(product, component, acr, projectBranch.imageTag(), env.GIT_COMMIT, env.LAST_COMMIT_TIMESTAMP)
-    onPR {
-      acr.retagForStage(DockerImage.DeploymentStage.PR, dockerImage)
+    if (fileExists('Dockerfile')) {
```
timja, line 44: this check should be inside onPR so we don't evaluate it outside of PRs

PR #929, opened 2022-11-16
Add Conditions to after for better display and reporting

timja:
Fixes a few issues:

1. Errors in after callbacks currently prevent metrics being collected because they are executed in a finally block before the metrics
2. Some builds run their integration tests in an after callback, we've seen cases where compilation has failed or unit tests have failed and then the integration tests run anyway, causing confusion and misreporting in metrics. These should be swapped to `afterSuccess`.
3. (similar to 2) General confusion that after block runs even when build failed, and so you get an error at the bottom of your build log when the actual build cause was further up.

This introduces 3 conditions that you can use to select when you want the callbacks to run:

- failure
- sucess
- always

I've also deprecated the default after so we don't have these reporting issues forever.

Tested in https://github.com/hmcts/draft-store/pull/1150

image

### DIFF
--- a/src/uk/gov/hmcts/contino/PipelineCallbacksRunner.groovy
+++ b/src/uk/gov/hmcts/contino/PipelineCallbacksRunner.groovy
@@ -1,5 +1,9 @@
 package uk.gov.hmcts.contino
 
+import uk.gov.hmcts.pipeline.deprecation.WarningCollector
+
+import java.time.LocalDate
+
 class PipelineCallbacksRunner implements Serializable {
   final PipelineCallbacksConfig config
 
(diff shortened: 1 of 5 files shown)


src/uk/gov/hmcts/contino/PipelineCallbacksRunner.groovy
```diff
@@ -1,38 +1,70 @@
...
+        callAfterAlways(stage)
+      } catch (err) {
+        call('onStageFailure', stage)
+        errToThrow = err
+      }
       nullSafeCall('after:all', stage)
-    }
-  }
-
-  void callAround(String stage, boolean condition, Closure body) {
```
timja, line 31: this was unused and would never have worked as `echo` is not in scope here

adusumillipraveen (approved):

PR #959, opened 2023-01-06
Save pods logs in artifacts at the end of tests

filippo-possenti-version1:
Notes:
In Civil we often hit the pods limit. This happens because failing builds keep the pods running to give the developer a chance to look into their logs and unfortunately our builds fail very often for various reasons.

This PR is part of a set of changes that aim at removing the pods regardless of success or failure.

This particular PR focuses on saving the pods logs into Jenkins artifacts so that the developer will no longer need the pods running to check their related logs.

For reference to anyone who is going to review, this is what you should see in Jenkins:
image

And this is a log example:
image

Based on this PR:
https://build.platform.hmcts.net/blue/organizations/jenkins/HMCTS_a_to_c%2Fcivil-ccd-definition/detail/PR-1990/28/artifacts/

PR #963, opened 2023-01-11
Add support to delete helm releases on failure

filippo-possenti-version1:
In Civil we often hit the pods limit. This happens because failing builds keep the pods running to give the developer a chance to look into their logs and unfortunately our builds fail very often for various reasons.

This PR is part of a set of changes that aim at removing the pods regardless of success or failure.

This particular PR focuses on providing functionality that will allow to delete the helm releases also for failing builds and it's meant to be tested and released after #959 (it touches the same areas and the changes are expected to be functionally conflicting). Other PRs in the civil namespace will then use this functionality to clear the pods.

The expectation is that this change will enable the Civil teams to rid themselves of a problem that is all too often stopping development and testing activity and may even result in a significant reduction in the overall amount of used resources, which may allow the Civil teams to start tackling other related issues.

The following is an example of us running out of pods:
image

After the changes in this PR, adding the line highlighted below will allow to delete the helm releases even after failure:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,35 @@ To clear helm release, do the following:
   }
   ```
 
+#### Clear Helm Release on Failed Build
+
+If your team doesn't need to keep the helm releases for their PR even for failing builds, it is possible to clean them to free further resources.
+This may be useful to reduce resource contention in difficult projects with large teams, where pipeline stability is not yet where it should be.
+
+To clear helm releases for failing builds, do the following:
+
+  ```
+  withPipeline(type, product, component) {
+    ...
+    enableCleanupOfHelmReleaseOnFail()
+    ...
+  }
+  ```
+
+If you know you want to keep the helm releases for specific failures, you can do the following:
+
+  ```
+  withPipeline(type, product, component) {
+    ...
+    disableCleanupOfHelmReleaseOnFail()
+    ...
+  }
+  ```
+
+Note that the above works only for PR builds. Master builds are unaffected by this toggle.
+
+
+
 ### Opinionated infrastructure pipeline
 
 For infrastructure-only repositories e.g. "shared infrastructure" the library provides an opinionated infrastructure pipeline which will build Terraform files in the root of the repository.
--- a/README.md
+++ b/README.md
@@ -266,6 +266,8 @@ The API tests run after smoke tests.
 If your server never uses the deployed resources, teams can clear the helm release to free resources on the cluster.
 This can be done always or depending on success.
 
+To note that clearing the helm release won't deny you access to the pods logs, as they are saved before the helm release is cleared.
+
 To clear the helm release regardless of success or failure, do the following:
 
   ```
@@ -295,8 +297,6 @@ To clear the helm release on a failing build, do the following:
     ...
   }
   ```
-
-To note that clearing the helm release won't deny you access to the pods logs, as they are saved before the helm release is cleared.
 
 
 ### Opinionated infrastructure pipeline
--- a/README.md
+++ b/README.md
@@ -261,12 +261,22 @@ tests for that API. For the pipeline to run those tests, do the following:
 
 The API tests run after smoke tests.
 
+#### Clear Helm Release
 
-#### Clear Helm Release on Successful Build
+If your server never uses the deployed resources, teams can clear the helm release to free resources on the cluster.
+This can be done always or depending on success.
 
-If your service never use the deployed resources once the build is green, teams can clear the helm release to free resources on the cluster.
+To clear the helm release regardless of success or failure, do the following:
 
-To clear helm release, do the following:
+  ```
+  withPipeline(type, product, component) {
+    ...
+    enableCleanupOfHelmReleaseAlways()
+    ...
+  }
+  ```
+
+To clear the helm release on a successful build, do the following:
 
   ```
   withPipeline(type, product, component) {
@@ -276,12 +286,7 @@ To clear helm release, do the following:
   }
   ```
 
-#### Clear Helm Release on Failed Build
-
-If your team doesn't need to keep the helm releases for their PR even for failing builds, it is possible to clean them to free further resources.
-This may be useful to reduce resource contention in difficult projects with large teams, where pipeline stability is not yet where it should be.
-
-To clear helm releases for failing builds, do the following:
+To clear the helm release on a failing build, do the following:
 
   ```
   withPipeline(type, product, component) {
@@ -291,21 +296,7 @@ To clear helm releases for failing builds, do the following:
   }
   ```
 
-#### Clear Helm Release always
-
-If your team doesn't need to keep the helm releases for their PR regardless of success or failure, it is possible to clean them.
-
-To clear helm releases regardless of success, do the following:
-
-  ```
-  withPipeline(type, product, component) {
-    ...
-    enableCleanupOfHelmReleaseOnFailure()
-    ...
-  }
-  ```
-
-This essentially combines `enableCleanupOfHelmReleaseOnSuccess` and `enableCleanupOfHelmReleaseOnFailure` in one command.
+To note that clearing the helm release won't deny you access to the pods logs, as they are saved before the helm release is cleared.
 
 
 ### Opinionated infrastructure pipeline
--- a/README.md
+++ b/README.md
@@ -291,6 +291,22 @@ To clear helm releases for failing builds, do the following:
   }
   ```
 
+#### Clear Helm Release always
+
+If your team doesn't need to keep the helm releases for their PR regardless of success or failure, it is possible to clean them.
+
+To clear helm releases regardless of success, do the following:
+
+  ```
+  withPipeline(type, product, component) {
+    ...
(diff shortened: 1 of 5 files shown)


vars/sectionDeployToAKS.groovy
```diff
@@ -94,8 +101,11 @@ def call(params) {
                 timeoutWithMsg(time: 10, unit: 'MINUTES', action: 'Smoke Test - AKS') {
                   try {
                     builder.smokeTest()
-                  } finally {
                     savePodsLogs(dockerImage, params, "smoke")
+                  } catch(err) {
+                    savePodsLogs(dockerImage, params, "smoke")
+                    clearHelmReleaseForFailure(config, dockerImage, params, pcr)
+                    error 'Build failed. Please check the build logs and the pods logs for further information. You will find the pods logs under Artifacts.'
```
adusumillipraveen, line 108: Should we log this and throw actual err Thoughts @timja ?

filippo-possenti-version1: I tried with "throw err"... didn't seem to make much of a difference in the logs and I found this to be clearer.

The commit history should show these attempts, I believe.

timja: I think this could hide the actual error though especially if something inside is throwing a specific error then that will never be shown I think with this

filippo-possenti-version1: That was my initial thought as well, which is why I tried with "throw err", only to realize the returned error in that case was just a numerical code with no particular value and everything else was identical.

I do however appreciate that this might have been specific to my build, so I'll put "throw err" back in, as from development perspective I also consider it the better option.

vars/sectionDeployToAKS.groovy
```diff
@@ -111,8 +121,11 @@ def call(params) {
                       timeoutWithMsg(time: config.fullFunctionalTestTimeout, unit: 'MINUTES', action: 'Functional tests') {
                         try {
                           builder.fullFunctionalTest()
-                        } finally {
                           savePodsLogs(dockerImage, params, "full-functional")
```
adusumillipraveen, line 124: Isn't this duplicate?

filippo-possenti-version1: Not sure what you mean there.

Before this change, when I added the pods logs, I added them in the `finally` clause as I wanted to have those logs regardless.
I still want that of course, but I can't use the `finally` block any longer as if I were to delete the helm release as part of the `catch` block, by the time I got to the `finally` the logs would be gone.

As a result I had to get rid of the `finally` block and explicitly put `savePodsLogs` both in the `try` and `catch` blocks, with the caveat being that the one in the `catch` block is executed BEFORE I clear the helm release.

Does it make sense or am I misunderstanding your concern?

timja: i don't think we want to save them twice. 

best to set a boolean value that tells the finally block the status

https://stackoverflow.com/a/10736283/4951015

```groovy
success = true
try {
 ...
} catch(err) {
  success = false
} finally {
  savePodLogs(...)
  if (!success) { 
    clearHelmReleaseForFailure(config, dockerImage, params, pcr)
  }
}
```

README.md
```diff
@@ -276,6 +276,22 @@ To clear helm release, do the following:
   }
   ```
 
+#### Clear Helm Release on Failed Build
```
timja, line 279: is there a reason we wouldn't want to do this by default given we're capturing the pods logs? I guess if you wanted to check the pod environment variables etc? but we should capture the pod manifest as well maybe ?

adusumillipraveen (dismissed):

README.md
```diff
@@ -276,6 +276,22 @@ To clear helm release, do the following:
...
+
+If your team doesn't need to keep the helm releases for their PR even for failing builds, it is possible to clean them to free further resources.
+This may be useful to reduce resource contention in difficult projects with large teams, where pipeline stability is not yet where it should be.
+
+To clear helm releases for failing builds, do the following:
+
+  ```
+  withPipeline(type, product, component) {
+    ...
+    enableCleanupOfHelmReleaseOnFail()
```
timja, line 289: It would be good if there was a way so that you don't need to add two lines to cleanup success and failure, i.e. in your screenshot you're having to add two lines of config

You could add something like this as well?
```
enableCleanupOfHelmReleaseAlways()
```

That would map to the existing callback names?
https://github.com/hmcts/cnp-jenkins-library#extending-the-opinionated-pipeline

🚫 At a minimum I would suggest renaming this to match the callbacks:

```suggestion
    enableCleanupOfHelmReleaseOnFailure()
```

timja (changes requested):
Close but I think needs a bit of cleanup and consistency changes

vars/sectionDeployToAKS.groovy
filippo-possenti-version1: I'd like to stress that we won't save them twice.

Anyway: in the interest of speed and since I don't own this repository, I'll follow your pattern.

README.md
filippo-possenti-version1: The reason is simply that I don't presume to know how other teams like to work.

For now it's just safer to add the feature as an "opt-in". Then platops can do further considerations as to the opportunity to enable this by default (which I would imagine would be the best thing to do) and what other changes are needed to support that properly.

README.md
filippo-possenti-version1: I'll match the callback names.

As for the additional "always" bit, I would leave further simplifications to your team. The purpose of this PR from my standpoint is simply to finally overcome an issue that is frequently blocking us in civil.

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -78,6 +78,15 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.clearHelmRelease = true;
   }
 
+  void enableCleanupOfHelmReleaseOnFailure() {
+    config.clearHelmReleaseOnFailure = true;
+  }
+
+  void enableCleanupOfHelmReleaseAlways() {
+    config.clearHelmRelease = true;
```
adusumillipraveen, line 86: Should we rename it ?
```suggestion
    config.clearHelmReleaseOnSuccess = true;
```

filippo-possenti-version1: I don't want to change things that were already there, if at all possible. Not as part of this ticket anyway.

I'd like this ticket to be laser-focused to solving the issue at hand. Once that's done, I'll be more than happy to do some minor improvements here and there whenever I get the time.

adusumillipraveen: Hmm, I don't think it's unrelated. We added a new var `clearHelmReleaseOnFailure` , so it makes sense to refactor existing var to make it more meaningful. Nothing should break for changing that.

filippo-possenti-version1: Not unrelated, just not necessary given the scope of this ticket. I merely preferred to separate the feature from the refactoring, as it keeps the PR smaller and therefore easier to understand.

Anyway: alright. Changed it there and in all places where it's used in this repo.

README.md
```diff
@@ -276,6 +286,19 @@ To clear helm release, do the following:
...
+
+  ```
+  withPipeline(type, product, component) {
+    ...
+    enableCleanupOfHelmReleaseOnFailure()
+    ...
+  }
+  ```
+
+To note that clearing the helm release won't deny you access to the pods logs, as they are saved before the helm release is cleared.
```
adusumillipraveen:
```suggestion
```

README.md
```diff
@@ -261,12 +261,22 @@ tests for that API. For the pipeline to run those tests, do the following:
 
 The API tests run after smoke tests.
 
+#### Clear Helm Release
 
-#### Clear Helm Release on Successful Build
+If your server never uses the deployed resources, teams can clear the helm release to free resources on the cluster.
```
adusumillipraveen, line 266:
```suggestion
If your service never uses the deployed resources after the pipeline is successful or if you refer to them only for logs in case of failure, you can uninstall the helm release to free resources on the cluster. Note that clearing the helm release won't deny you access to the pods logs, as they are saved as artefacts in Jenkins before the helm release is cleared.
```

README.md
```diff
@@ -261,12 +261,22 @@ tests for that API. For the pipeline to run those tests, do the following:
...
+This can be done always or depending on success.
```
adusumillipraveen, line 267:
```suggestion
```

README.md
```diff
@@ -261,12 +261,24 @@ tests for that API. For the pipeline to run those tests, do the following:
...
 
-If your service never use the deployed resources once the build is green, teams can clear the helm release to free resources on the cluster.
+To note that clearing the helm release won't deny you access to the pods logs, as they are saved before the helm release is cleared.
```
adusumillipraveen, line 269:
```suggestion
```

adusumillipraveen (approved):
Nice work 🎉

PR #978, opened 2023-01-24
Add pipeline nagger for azurerm 3.x and terraform 1.x

timja:
Required so we can merge https://github.com/hmcts/cnp-module-redis/pull/33

Many teams aren't updating azurerm versions, (more teams have been updating terraform versions for some reason though)

image

Tested in https://github.com/hmcts/fis-shared-infrastructure/pull/14


vars/spinInfra.groovy
```diff
@@ -91,8 +91,6 @@ def call(Map<String, ?> params) {
 
         sh "terraform --version"
 
-        warnAboutOldTfAzureProvider()
```
timja, line 94: I don't think this would have worked as intended before, the provider version doesn't show up until you've done a terraform init, but could well show up while you are testing if you've previously done an init...

adusumillipraveen (approved):

adusumillipraveen (approved):

thomast1906 (approved):

PR #995, opened 2023-02-04
set +x to prevent command echoing

### LINKED ISSUE #994: Credentials leaked via pipeline on all Java projects
### What would you like to change?
Credentials are currently leaked to all pipeline users on the "Security Check" stage on any build that utilises the GradleBuilder.
This should probably not happen?

### How do you think that would improve the project?
Security would be improved.

### If this entry is related to a bug, please provide the steps to reproduce it
1) Go to any java build within our pipeline.
2) Go to the security check step
3) Read the shell output from the third dropdown - image below:
 (image

furnivall:
Resolves #994 . (This is applicable only if this pull request relates to a GitHub issue, delete the line otherwise)

Notes:
* hidden the shell execution of the ./gradlew step
* moved away from the gradle wrapper function for this (to allow the shell to be intercepted by tests and to use +x for an individual command)
* There is a high likelihood the spock test is not great. If a fix is needed let me know.

### DIFF
--- a/src/uk/gov/hmcts/contino/GradleBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/GradleBuilder.groovy
@@ -136,7 +136,9 @@ class GradleBuilder extends AbstractBuilder {
     ]
     localSteps.withAzureKeyvault(secrets) {
       try {
+          gradle("set +x")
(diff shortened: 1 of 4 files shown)


timja (changes requested):
we want command echoing, this should instead use single quotes instead of double quotes

src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.contino
 
 import groovy.json.JsonSlurper
+import hudson.Functions
```
timja, line 4:
```suggestion
```

PR #1014, opened 2023-03-02
DTSPO-12454 Use java 17 for running gradle when project requests 17

timja:
https://tools.hmcts.net/jira/browse/DTSPO-12454

Spring Boot 3 gradle plugin requires Java 17
Previously we used toolchains for this and ran on Java 11, but now that a Gradle plugin requires 17, the gradle jvm needs to be running 17 as well

### DIFF
--- a/resources/uk/gov/hmcts/gradle/init.gradle
+++ b/resources/uk/gov/hmcts/gradle/init.gradle
@@ -24,10 +24,6 @@ allprojects {
   task printVersionInit {
     doLast { println project.version }
   }
-
-  task javaVersion {
-    doLast { println project.sourceCompatibility }
-  }
 }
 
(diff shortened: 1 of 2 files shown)


resources/uk/gov/hmcts/gradle/init.gradle
```diff
@@ -24,10 +24,6 @@ allprojects {
   task printVersionInit {
     doLast { println project.version }
   }
-
-  task javaVersion {
```
timja, line 28: Turns out this isn't useful as we need to run on Java 17 to run gradle at all if you have a plugin that requires it.
Could run the gradle jvm on 17 just for this check but that may fail on old gradle versions

adusumillipraveen (approved):

PR #1027, opened 2023-03-31
Updated ccd chart version in check-deprecated-charts.sh

mokainos:

Notes:
*
*
*


msl8r (approved):

timja (changes requested):
we should update the expiry date when we do this 

https://github.com/hmcts/cnp-jenkins-library/blob/6ed3edce9155b5da48caac7a445879adfedfb70b/vars/warnAboutDeprecatedChartConfig.groovy#L28

PR #1033, opened 2023-04-04
CIV-7142 Ensure deletion on pods on failed deployment

filippo-possenti-version1:
Resolves https://tools.hmcts.net/jira/browse/CIV-7142.

While looking into the PRs created by renovate for civil, it was noticed that some of them were aborted due to deployment failures (the usual quota issue) and despite this, they were in fact deployed.

The result was that for each of these automated PRs experiencing the quota issue we had 10 running pods, thus further compounding to the quota issues for all subsequent builds. 3 PRs were identified with this issue in just one day, leading to believe the issue could be quite common whenever the quota limits are reached, making things worse exactly at the worst possible time.

This change aims at addressing the issue by clearing the helm release when the deployment has failed and the `clearHelmReleaseOnFailure` flag is set.


timja (approved):
LGTM

PR #1056, opened 2023-04-26
DTSPO-12678 Add workload identity nagger to pipeline

reespozzi:
Update old file to search instead for workload identity flag being set to true (teams have migrated)

Notes:
*
*
*


adusumillipraveen (approved):
Do not merge till we finish implementation

reespozzi:
If `aadIdentityName` flag is removed from application code, pipeline runs with no issues as `useWorkloadIdentity` flag is now set in latest base charts.

image


If aadidentityname flag is still present, migration to workload identity may not be fully complete, and the pipeline will be blocking (After the decided date)
image

resources/uk/gov/hmcts/helm/warn-about-workload-identity.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+CHART_DIRECTORY=${1}-${2}
+cd charts/"${CHART_DIRECTORY}" || exit 10
+# Check if app still has aad config
+grep --quiet  "aadIdentityName:" values.yaml
```
adusumillipraveen, line 6: Better to check in values.*.yaml ? like preview and aat?

reespozzi: Yup good idea, I added to search all .yaml files in general in case they're named differently

adusumillipraveen (approved):

PR #1070, opened 2023-06-02
DTSPO-13832 - add check for java 17 in dbMigrate

endakelly:
dbMigrate stage uses default java installation on agent which is still version 11. This is causing an issue on draft-store as some recently updated dependencies require java 17.

### DIFF
--- a/src/uk/gov/hmcts/contino/GradleBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/GradleBuilder.groovy
@@ -222,35 +222,6 @@ EOF
       functionalTest()
   }
 
-  def dbMigrate(String vaultName, String microserviceName) {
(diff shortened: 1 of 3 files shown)


timja (changes requested):
don't duplicate the code you probably just want to setup tool versions earlier?

see this PR:
https://github.com/hmcts/cnp-jenkins-library/pull/1064

endakelly:
Is [this](https://github.com/hmcts/cnp-jenkins-library/pull/1070/files#diff-512c3876f69aa558402364287f8712720bb99311d02e7750857331fe998882a0R34) more like it? There's not an exact parallel with the camunda pipeline in withPipeline so took the cue from sectionBuildAndTest

PR #1108, opened 2023-07-24
DTSPO-12193 - Remove aksClient

endakelly:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/DTSPO-12193

### Change description ###
Remove aksClient
Jenkins agent now has tools preinstalled

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

### DIFF
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -96,9 +96,8 @@ def call(params) {
                 writeFile file: '.dockerignore_build', text: libraryResource('uk/gov/hmcts/.dockerignore_build')
                 sh script: """
                         # in case anyone doesn't have a trailing new line in their file
-                        echo -e '\n' >> .dockerignore
+                        printf '\r\n' >> .dockerignore
                         cat .dockerignore_build >> .dockerignore
-                        sed -i 's/-e//g' .dockerignore
                       """
               }
               def buildArgs = projectBranch.isPR() ? " --build-arg DEV_MODE=true" : ""
--- a/vars/sectionBuildAndTest.groovy
+++ b/vars/sectionBuildAndTest.groovy
@@ -98,6 +98,7 @@ def call(params) {
                         # in case anyone doesn't have a trailing new line in their file
                         echo -e '\n' >> .dockerignore
                         cat .dockerignore_build >> .dockerignore
+                        sed -i 's/-e//g' .dockerignore
                       """
               }
               def buildArgs = projectBranch.isPR() ? " --build-arg DEV_MODE=true" : ""
--- a/vars/withAcrClient.groovy
+++ b/vars/withAcrClient.groovy
@@ -1,6 +1,6 @@
 
 def call(String subscription, Closure block) {
-  withSubscriptionLogin(subscription, alwaysLogin) {
+  withSubscriptionLogin(subscription) {
     withRegistrySecrets {
       block.call()
     }
--- a/vars/withAcrClient.groovy
+++ b/vars/withAcrClient.groovy
@@ -1,11 +1,7 @@
 
 def call(String subscription, Closure block) {
-  if (env.IS_DOCKER_BUILD_AGENT && env.IS_DOCKER_BUILD_AGENT.toBoolean()) {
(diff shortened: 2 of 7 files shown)


timja (changes requested):
the whole point of this is to get rid of the withDocker call here:
https://github.com/hmcts/cnp-jenkins-library/blob/d841760cfc3a8a0def31ed7405a663ef011b60ff/vars/withAcrClient.groovy#L6

you may want to keep the withAksClient so that the cluster variables get set properly.

basically the docker image is no longer needed because we install all tools on the packer agent now

vars/withAcrClient.groovy
```diff
@@ -3,9 +3,7 @@ def call(String subscription, Closure block) {
   if (env.IS_DOCKER_BUILD_AGENT && env.IS_DOCKER_BUILD_AGENT.toBoolean()) {
     doAcrOrKubectlTask(subscription, false, block)
   } else {
-    withDocker('hmcts/cnp-aks-client:az-2.44.1-kubectl-1.26.0-helm-3.10.3', null) {
       doAcrOrKubectlTask(subscription, true, block)
```
timja, line 6: also no need to set this to true anymore as it's no longer in a container.

you should be able to remove `doAcrOrKubectlTask` as it's no longer needed, and might be able to remove the `alwaysLogin` functionality if it's no longer needed

timja (changes requested):
close

vars/withAcrClient.groovy
endakelly: Something like this? https://github.com/hmcts/cnp-jenkins-library/pull/1108/files#diff-5348ea2ed32827d22070457c3804a837f4cbbdd885a6e2d46a2939cb135294dcR2-R5

timja: almost, https://github.com/hmcts/cnp-jenkins-library/blob/d841760cfc3a8a0def31ed7405a663ef011b60ff/vars/withSubscriptionLogin.groovy#L8-L15 is no longer used so should be removed

vars/sectionBuildAndTest.groovy
```diff
@@ -98,6 +98,7 @@ def call(params) {
                         # in case anyone doesn't have a trailing new line in their file
                         echo -e '\n' >> .dockerignore
                         cat .dockerignore_build >> .dockerignore
+                        sed -i 's/-e//g' .dockerignore
```
timja, line 101: what's this for?

endakelly: For some reason the docker ignore is appending `-e` to the `.jar` file. I will change it to replace `.jar-e` with `.jar` so it doesn't muck up images with `-e` in their names. I think it's something to do with the echo

timja: probably needs a bit more debugging, seems quite odd =/

timja (changes requested):

vars/withSubscriptionLogin.groovy
```diff
@@ -5,28 +5,19 @@
  * @param body the body to execute after logged in
  */
 def call(String subscription, boolean alwaysExecute, Closure body) {
```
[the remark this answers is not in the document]
timja: still present

timja (changes requested):
https://github.com/hmcts/cnp-jenkins-library/pull/1108/files/d7ccceb12f1585b25738ffecfcd7fd4e44552da0#diff-1d58eed88e02bb1f6c54e9e4799c78f65c2d05a79afa269c785a1a0e135d5e94

timja (approved):

vars/withSubscriptionLogin.groovy
```diff
@@ -32,4 +32,4 @@ def call(String subscription, boolean alwaysExecute, Closure body) {
 
 def call(String subscription, Closure body) {
```
endakelly, line 33: @timja the pipeline is complaining about this being a duplicate method of the same signature. Can this block be removed or should it be renamed?

timja: remove it as it's no longer needed

timja (approved):

PR #1111, opened 2023-07-31
Don't access deprecation config static var

timja:
User reported null helm object after a Jenkins restart.

Looking at the code there's no need for us to not call load every time as it caches a copy


adusumillipraveen (approved):

vars/spinInfra.groovy
```diff
@@ -60,12 +60,16 @@ def call(Map<String, ?> params) {
...
         def builtFrom = env.GIT_URL ?: 'unknown'
 
-        def tags = [environment: Environment.toTagName(config.environment), changeUrl: changeUrl, managedBy: teamName, BuiltFrom: builtFrom, contactSlackChannel: contactSlackChannel, application: env.TEAM_APPLICATION_TAG, businessArea: env.BUSINESS_AREA_TAG]
+        def tags = [environment: Environment.toTagName(config.environment), managedBy: teamName, builtFrom: builtFrom, contactSlackChannel: contactSlackChannel, application: env.TEAM_APPLICATION_TAG, businessArea: env.BUSINESS_AREA_TAG]
 
         if (Environment.toTagName(config.environment) == "sandbox") {
           tags = tags + [expiresAfter: config.expires]
         }
 
+        if (changeUrl && changeUrl != "null" && changeUrl != "") {
```
timja, line 69: I got a really strange error about `null` being a reserved word and not allowed as a tag value, it was trying to change from empty string to `null`, prevent all that as it's not needed / wanted.

adusumillipraveen (approved):

PR #1130, opened 2023-08-11
DTSSE-2204/ardoq integration prototype

jasonpaige:
This change introduces a new optional build step. It currently only works for yarn projects but the intent is to enable gradle very shortly.
Teams will be onboarded via the jenkins config like so: https://github.com/hmcts/cnp-jenkins-config/blob/master/team-config.yml#L572

This should therefor be a non-breaking change.

Currently testing this here: https://github.com/hmcts/cnp-jenkins-config/blob/master/team-config.yml#L572


src/uk/gov/hmcts/ardoq/ArdoqClient.groovy
```diff
@@ -0,0 +1,87 @@
...
+      return
+    }
+    return applicationId?.trim();
+  }
+
+  String getRepositoryName() {
+    return new RepositoryUrl().getShortWithoutOrgOrSuffix(this.steps.env.GIT_URL)
+  }
+
+  String getLanguage() {
```
timja, line 29: might be better to pass this in or subclass per language rather than trying to detect as it should probably be unneeded? but no harm for now

timja (approved):
lgtm

src/uk/gov/hmcts/ardoq/ArdoqClient.groovy
jasonpaige: I think in order to get the specific version this seemed like the most reliable way and just happens to work for both Java and Nodejs.
You're right though, as this grows to other ares of the business to support non-dockerised repos I'll need to refactor this.

PR #1148, opened 2023-09-06
DTSPO-15007 - add nagger for postgres module

endakelly:
Add nagger to tell teams that cnp-module-postgres is deprecated

image

### DIFF
--- a/resources/uk/gov/hmcts/infrastructure/check-deprecated-postgres.sh
+++ b/resources/uk/gov/hmcts/infrastructure/check-deprecated-postgres.sh
@@ -2,7 +2,7 @@
 
 grep -R --quiet "hmcts/cnp-module-postgres" *
 
-if grep -R --quiet "hmcts/cnp-module-postgres" *; then
+if grep -R --quiet "hmcts/cnp-module-postgres"; then
   echo "====================================================================================================="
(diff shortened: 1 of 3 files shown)


resources/uk/gov/hmcts/infrastructure/check-deprecated-postgres.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+output=$(grep -R --include=*.tf "hmcts/cnp-module-postgres" | wc -l)
+
+if [ $(echo $output | tr -d " ") = 0 ]; then
+  echo "Not using cnp-module-postgres module, this is good"
```
timja, line 6: probably don't need to echo anything in this case

endakelly: It doesn't actually output anything. Initially I tried something like [this](https://github.com/hmcts/cnp-jenkins-library/blob/9d751c190da3bc9b6d8326afc0cacb3ad20577c5/resources/uk/gov/hmcts/helm/warn-about-aad-identity-preview-hack.sh#L6-L15) but it wasn't giving me consistent results for some reason

timja (approved):

PR #1159, opened 2023-09-25
Extend a teams renovate exception by one week

thomast1906:
Extend a teams renovate exception by one week


timja (changes requested):
this is a 30 second change, no

PR #1165, opened 2023-10-11
DTSPO-15437 - Add optional parameter to run frontend security script

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-15437

Enables teams like IDAM who use gradle for their frontend applications to use our central security script


endakelly:
@adusumillipraveen @timja could you advise on this one? I have managed to get the pipeline working when I've got two parameters passed to `enableSecurityScan` e.g. https://github.com/hmcts/cnp-plum-recipes-service/blob/4b491346ced36a30730847d424c2e8c51dcd7e43/Jenkinsfile_nightly#L26 and https://github.com/hmcts/idam-web-public/blob/865c1b97ee5b373aaf213e36711b721b44ccefeb/Jenkinsfile_nightly#L69 but I'm struggling to figure out how to make this optional as it doesn't work when nothing is passed or only one parameter is passed e.g. https://github.com/hmcts/cnp-plum-frontend/blob/9e0a32e29cacdc2675c1752254a122d7f6dbf08e/Jenkinsfile_nightly#L19

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -43,8 +43,9 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.parallelCrossBrowsers = browsers
   }
 
-  void enableSecurityScan(String urlExclusions = "", int timeout = 120) {
+  void enableSecurityScan(String urlExclusions = "", isFrontend = false, int timeout = 120) {
```
timja, line 46: instead of using positional arguments you could use a map:

```suggestion
  void enableSecurityScan(Map<String, String> args = [:]) {
        def config = [
            urlExclusions: "",
            isFrontend: false,
            timeout: 120
    ] << args
  }
```

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -43,10 +43,16 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
     config.parallelCrossBrowsers = browsers
   }
 
-  void enableSecurityScan(String urlExclusions = "", int timeout = 120) {
-    config.urlExclusions = urlExclusions
-    config.securityScanTimeout = timeout
-    config.securityScan = true
+  void enableSecurityScan(Map<String, String> args = [:]) {
```
timja, line 46:
```suggestion
  void enableSecurityScan(Map<String, Object> args = [:]) {
```

README.md
```diff
@@ -434,9 +434,17 @@ You can find an example in [idam-web-public](https://github.com/hmcts/idam-web-p
...
-The pipeline will automatically detect whether you're application is node based or gradle based and run the appropriate security tests based on that.
+The pipeline will automatically detect whether your application is node based or gradle based and run the appropriate security tests based on that.
 
-If you have a requirement to customise the script, you can place your own script in a folder called `ci` in your repo. Make sure to call the script `security.sh`.
+Gradle based applications are more commonly used in the backend but if your frontend application is gradle based, you can pass a parameter to `withNightlyPipeline` to indicate this is the case and run the frontend specific security script instead of the default backend specific script. This value is optional.
+
+```
+def isFrontend = true
+
+withNightlyPipeline(type, product, component, isFrontend)
```
timja, line 444: no we shouldn't modify the parameters, they are fixed for all pipelines at:
`type, product, component`

in the DSL callbacks a flag can be added, something like:

```
withNightlyPipeline(type, product, component) {
  useFrontendSecurityTests()
}
```

timja (changes requested):

README.md
endakelly: Would that be a case of adding a new function to the dsl like this?
```
  void useFrontendSecurityTests(String urlExclusions = "", int timeout = 120) {
    config.urlExclusions = urlExclusions
    config.securityScanTimeout = timeout
    config.useFrontendSecurityTests = true
  }
```

And then new logic to `sectionNightlyTests` where the variable would be set as required?
```
    if (config.useFrontendSecurityTests) {
      stageWithAgent('Security scan', product) {
        env.IS_FRONTEND = "true"
        warnError('Failure in securityScan') {
          pcr.callAround('securityScan') {
            timeout(time: config.securityScanTimeout, unit: 'MINUTES') {
              builder.securityScan()
            }
          }
        }
      }
    }
```

timja: um no not really, we can pair tomorrow to finish this off

test/uk/gov/hmcts/contino/AppPipelineConfigTest.groovy
```diff
@@ -42,9 +42,10 @@ class AppPipelineConfigTest extends Specification {
     when:
       dsl.enableSecurityScan()
     then:
-      assertThat(pipelineConfig.securityScan).isTrue()
+      assertThat(pipelineConfig.securityScanUrlExclusions).isEqualTo("")
+      assertThat(pipelineConfig.useFrontendSecurityScan).isEqualTo("false")
```
endakelly, line 46: Not sure why it's expecting this to be false - is this even really needed here?

timja: I would expect it to be this:
```suggestion
      assertThat(pipelineConfig.useFrontendSecurityScan).isFalse()
```

endakelly: I thought the same but it seems that it's a string rather than a boolean. Setting it to `isFalse()` causes an error

endakelly: I think I've fixed it. Updated the configuration.scanType to use params.scanType if present or default to `""` and updated the GradleBuilder to expect the value of the environment variable to be `frontend`. Previously was set to expect `true` when I was using a boolean value.

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -43,9 +43,16 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
...
-    config.securityScanTimeout = timeout
+  void enableSecurityScan(Map<String, Object> params = [:]) {
+    def configuration = [
+        urlExclusions: "",
+        timeout: 120,
+        scanType: params.scanType ? params.scanType : ""
+    ] << params
+
+    config.securityScanUrlExclusions = configuration.urlExclusions
+    config.useFrontendSecurityScan = configuration.scanType
```
timja, line 54: you will also need to change a few other places
```suggestion
    config.securityScanType = configuration.scanType
```

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
```diff
@@ -43,9 +43,16 @@ class AppPipelineDsl extends CommonPipelineDsl implements Serializable {
...
   }
 
-  void enableSecurityScan(String urlExclusions = "", int timeout = 120) {
-    config.urlExclusions = urlExclusions
-    config.securityScanTimeout = timeout
+  void enableSecurityScan(Map<String, Object> params = [:]) {
+    def configuration = [
+        urlExclusions: "",
+        timeout: 120,
+        scanType: params.scanType ? params.scanType : ""
```
timja, line 50:
```suggestion
        scanType: "auto"
```

timja (commented):
nice pretty close

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
endakelly: Updated

src/uk/gov/hmcts/contino/AppPipelineDsl.groovy
endakelly: Updated

timja (approved):
lgtm

PR #1174, opened 2023-10-30
DTSPO-14559 - Add logic for using CNAME instead of A record

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-14559

Adding logic and config to check private DNS zone for a cname record
If such a record exists, then create an A record pointing to the applications internal IP will not be created

Should also address https://hmcts-reform.slack.com/archives/C8SR5CAMU/p1693987408788219

Work pending to allow external-dns to exclude certain ingresses


src/uk/gov/hmcts/contino/AzPublicDns.groovy
```diff
@@ -0,0 +1,57 @@
...
+      this.environmentDnsConfigEntry = environmentDnsConfigEntry
+    }
+
+   def getHostName(recordName) {
+     def zone = this.environmentDnsConfigEntry.zone
+
+     return "${recordName}.${zone}"
+   }
+
+    def registerDns(recordName, serviceIP) {
```
timja, line 25: this isn't registering dns it's checking if a record-set already exists

timja (commented):
I think this might be a bit overly complex.
We shouldn't need to sync any records I think.

if someone wants to set this up they should just point the app to frontdoor in private dns then it should all work as long as Jenkins doesn't try updating it?

src/uk/gov/hmcts/contino/AzPublicDns.groovy
endakelly: Renamed it

endakelly:
> I think this might be a bit overly complex. We shouldn't need to sync any records I think.
> 
> if someone wants to set this up they should just point the app to frontdoor in private dns then it should all work as long as Jenkins doesn't try updating it?

I suppose we could just tell it to check for a CNAME in private dns and not update to an A record if it finds one but that would require teams to create a CNAME in private dns manually using the repo - which could be added to the docs. How does that sound?

The syncing suggestion was in the ticket originally so that's why I went down that route

timja:
> > I think this might be a bit overly complex. We shouldn't need to sync any records I think.
> > if someone wants to set this up they should just point the app to frontdoor in private dns then it should all work as long as Jenkins doesn't try updating it?
> 
> I suppose we could just tell it to check for a CNAME in private dns and not update to an A record if it finds one but that would require teams to create a CNAME in private dns manually using the repo - which could be added to the docs. How does that sound?
> 
> The syncing suggestion was in the ticket originally so that's why I went down that route

Yeah that sounds good

src/uk/gov/hmcts/contino/AzPrivateDns.groovy
```diff
@@ -44,19 +44,29 @@ class AzPrivateDns {
...
         try {
           aRecordSet = this.az.az "network private-dns record-set a show -g ${resourceGroup} -z ${zone} -n ${recordName} --subscription ${subscription} -o tsv"
         } catch (e) {
         } // do nothing, record not found
-        if (!aRecordSet) {
-          this.az.az "network private-dns record-set a create -g ${resourceGroup} -z ${zone} -n ${recordName} --ttl ${ttl} --subscription ${subscription}"
-          this.az.az "network private-dns record-set a add-record -g ${resourceGroup} -z ${zone} -n ${recordName} -a ${serviceIP} --subscription ${subscription}"
-        } else {
+        try {
+          cnameRecordSet = this.az.az "network private-dns record-set cname show -g ${resourceGroup} -z ${zone} -n ${recordName} --subscription ${subscription} -o tsv"
```
timja, line 58: if the a recordset is found there's no point also checking for a cname.

vars/helmInstall.groovy
```diff
@@ -28,6 +29,7 @@ def call(DockerImage dockerImage, Map params) {
   def aksServiceName = dockerImage.getAksServiceName()
 
   EnvironmentDnsConfigEntry dnsConfigEntry = new EnvironmentDnsConfig(this).getEntry(params.environment, product, component)
+  AzPublicDns azPublicDns = new AzPublicDns(this, params.environment, dnsConfigEntry)
```
timja, line 32: not needed?

test/uk/gov/hmcts/contino/AzPrivateDnsTest.groovy
```diff
@@ -60,10 +61,11 @@ class AzPrivateDnsTest extends Specification {
     def environment = "sbox"
     def recordName = "rn"
     def ip = "4.3.2.1"
+    def cname = ""
 
     when:
       azPrivateDns = Spy(AzPrivateDns, constructorArgs:[steps, environment, new EnvironmentDnsConfig(steps).getEntry(environment, 'plum', 'recipes-service')])
-      azPrivateDns.registerDns(recordName, ip)
+      azPrivateDns.registerDns(recordName, ip, cname)
```
timja, line 68: I don't think any of these cname changes are needed anymore?

src/uk/gov/hmcts/contino/AzPrivateDns.groovy
endakelly: Without the check, if the A record doesn't exist and the CNAME record does, the pipeline would try to create an A record and it would fail since you can only have one or the other.

README.md
```diff
@@ -67,6 +67,14 @@ Branch | Environment
 `perftest` | `perftest`
 PR branch| `preview`
 
+#### Running tests through Azure Front Door
+
+If you want tests in AAT / Stg environments to run via Azure Front Door, you must add configuration for your application to front door. Have a look at the [HMCTS Way](https://hmcts.github.io/cloud-native-platform/path-to-live/front-door.html#front-door-configuration).
+
+Add a CNAME for your application that points to front door to [azure-private-dns](https://github.com/hmcts/azure-private-dns). See [example](https://github.com/hmcts/azure-private-dns/blob/a79d79eb8050440053b45f1a9663db85badf4973/environments/staging/aat-platform-hmcts-net.yml#L65-L67).
```
timja, line 74: not a great example as it should be a staging record (include -staging at the end I think it is), this might confuse people

src/uk/gov/hmcts/contino/AzPrivateDns.groovy
timja: you are checking if the a record exists first, if A exists you don't need to check for CNAME, which will save time, its important to keep this optimised as it runs a lot

timja: (unsure if you're still working on it but it's still not right, we should only do this check if the a record is missing)

timja (changes requested):
reviewing to get off my dashboard, need to fix https://github.com/hmcts/cnp-jenkins-library/pull/1174/files#r1386914582

src/uk/gov/hmcts/contino/AzPrivateDns.groovy
endakelly: I get you now. I've moved the check for cname under the if statement.

timja (approved):

PR #1179, opened 2023-11-09
add new tag startupMode

viktoriyabogdanova:
Notes:
* DTSPO-15113 - add new tag startupMode
*
*

### DIFF
--- a/vars/spinInfra.groovy
+++ b/vars/spinInfra.groovy
@@ -68,6 +68,9 @@ def call(Map<String, ?> params) {
         if (Environment.toTagName(config.environment) != "production") {
           tags = tags + [autoShutdown: "true"]
         }
+        if (Environment.toTagName(config.environment) != "production") {
+          tags = tags + [startupMode: "always"]
+        }
         if (changeUrl && changeUrl != "null" && changeUrl != "") {
           tags = tags + [changeUrl: changeUrl]
         }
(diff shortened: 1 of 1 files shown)


vars/spinInfra.groovy
```diff
@@ -68,6 +68,9 @@ def call(Map<String, ?> params) {
         if (Environment.toTagName(config.environment) != "production") {
           tags = tags + [autoShutdown: "true"]
         }
+        if (Environment.toTagName(config.environment) != "production") {
+          tags = tags + [startupMode: "always"]
```
timja, line 72: shouldn't this be adding onDemand to sandbox? and what are you doing about old resources is there going to be a script ran to backfill or will it default to always? (even then a script would be needed for the likes of ITHC wouldn't it for when we enable it there?)

viktoriyabogdanova: Weren't we supposed to make it default always and then overwrite it per env per resource?

timja: it should be set at an environment level I think.

timja: cc @adusumillipraveen

adusumillipraveen: yeah, let's have a chat @viktoriyabogdanova , i think `always` is kind of default if not added. We need the other onDemand one for sandbox/ithc. And yes, we may have to do a script anyway to add tags to existing resources when we roll this out

timja (approved):

PR #1184, opened 2023-11-27
Yarn v4 audit implementation

furnivall:
In progress - still needs to be tested.

Audit format has changed from v3 -> v4. 
v4.0.0 uses a different format from 4.0.1 onwards, so we're not going to handle it and force upgrade to 4.0.1+.

Handles unnecessary suppressions and other bits. Will update this description later.

### DIFF
--- a/resources/uk/gov/hmcts/pipeline/yarn/yarnv4audit.py
+++ b/resources/uk/gov/hmcts/pipeline/yarn/yarnv4audit.py
@@ -1,0 +1,161 @@
+import json
+from dataclasses import dataclass, asdict
+import subprocess
+import os
+
+@dataclass
+class Vulnerability:
+  name: str
+  id: str
+  severity: str
+  vulnerable_versions: str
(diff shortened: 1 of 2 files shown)


resources/uk/gov/hmcts/pipeline/yarn/yarnv4audit.py
```diff
@@ -0,0 +1,163 @@
...
+    return
+  if len(unsuppressed_vulnerabilities) > 0:
+    print_vulnerabilities(unsuppressed_vulnerabilities)
+    return
+  if len(unneeded_suppressions) > 0:
+    print_suppressions(unneeded_suppressions)
+
+def build_parent_json_for_cosmosDB(vulnerabilities, suppressions):
+  if os.environ.get('CI') != 'true':
+    print("Not running in CI, skipping parent JSON block")
```
timja, line 150: this doesn't look like it skips anything?

PR #1189, opened 2023-12-05
Update to use new owasp dependency database

reespozzi:
Updates to use new DB that was built with latest version of dependencycheck 9.0.2

Provided on opt-in basis where teams will need latest version to use the newest DB

### DIFF
--- a/src/uk/gov/hmcts/contino/GradleBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/GradleBuilder.groovy
@@ -230,9 +230,9 @@ EOF
 
   private String gradleWithOutput(String task) {
     addInitScript()
-    return localSteps.sh(script: "./gradlew --no-daemon --stacktrace --init-script init.gradle ${task}", returnStdout: true).trim()
+    localSteps.sh(script: "./gradlew --no-daemon --init-script init.gradle ${task}", returnStdout: true).trim()
   }
-  
+
   def fullFunctionalTest() {
       functionalTest()
   }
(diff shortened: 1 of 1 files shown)


src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -125,12 +125,21 @@ class GradleBuilder extends AbstractBuilder {
...
-      [ secretType: 'Secret', name: 'OWASPPostgresDb-v14-Connection-String', version: '', envVariable: 'OWASPDB_V14_CONNECTION_STRING' ]
+      [ secretType: 'Secret', name: 'OWASPPostgresDb-v14-Connection-String', version: '', envVariable: 'OWASPDB_V14_CONNECTION_STRING' ],
+      [ secretType: 'Secret', name: 'OWASPPostgresDb-v15-Account', version: '', envVariable: 'OWASPDB_V15_ACCOUNT' ],
+      [ secretType: 'Secret', name: 'OWASPPostgresDb-v15-Password', version: '', envVariable: 'OWASPDB_V15_PASSWORD' ],
+      [ secretType: 'Secret', name: 'OWASPPostgresDb-v15-Connection-String', version: '', envVariable: 'OWASPDB_V15_CONNECTION_STRING' ]
     ]
 
     localSteps.withAzureKeyvault(secrets) {
       try {
+        if (hasPlugin("org.owasp.dependencycheck.gradle.plugin:9")) {
```
timja, line 136: I think we probably want to default to v9 and do v8 via feature detection.

There are some failure scenarios where artifacts can't be found but `hasPlugin` will fail and you'll get weird looking log errors and people will point at dependency check having an issue but it's actually a problem in their build

reespozzi:
```
There are some failure scenarios where artifacts can't be found but hasPlugin will fail and you'll get weird looking log errors and people will point at dependency check having an issue but it's actually a problem in their build
```
Can you expand a bit on this? Not sure how we'd make make sure the security check only runs against the new DB for versions > 9, if not using `hasPlugin` in the first instance like https://github.com/hmcts/cnp-jenkins-library/commit/aad317724c3e734cf321b0ee92611b61133752ce

Is there some other feature detection in this jenkins-library that's available that may be more suited?

timja (approved):

PR #1190, opened 2023-12-11
DTSPO-15526 - Add yarn env vars

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-15526

Add yarn env vars to builder so packages are downloaded to `/opt/app/.yarn` which will ensure Neuvector doesn't scan thousands of files
Tested on a few different repos and `.yarn` directory is mostly not even created in the home directory. Any time it has been, it has been empty. `/opt/app/.yarn` has the zip files in it.
Files on the docker build stage are within the container and not on the host filesystem

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,7 +466,7 @@ Gradle based applications are more commonly used in the backend but if your fron
 
 ```
 withNightlyPipeline(type, product, component) {
-  enableSecurityScan(                
+  enableSecurityScan(
     scanType: "frontend"
   )
 }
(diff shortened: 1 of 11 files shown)


timja (commented):
why is this needed? I'm surprised it matters for the jenkins library.

endakelly:
> why is this needed? I'm surprised it matters for the jenkins library.

There's two ways to ensure the home directory doesn't get used from what I have read up. You can set the `globalFolder` in the yarnrc.yml or use an environment variable. It would involve more work to set it on every repo. I did think about maybe having a nagger to check if it was set in the yarnrc.yml and send a warning about it.

Not every pipeline as is will create the `.yarn` directory in home but this should hopefully prevent it from happening so we can re-enable neuvector scanning.

timja:
> > why is this needed? I'm surprised it matters for the jenkins library.
> 
> There's two ways to ensure the home directory doesn't get used from what I have read up. You can set the `globalFolder` in the yarnrc.yml or use an environment variable. It would involve more work to set it on every repo. I did think about maybe having a nagger to check if it was set in the yarnrc.yml and send a warning about it.
> 
> Not every pipeline as is will create the `.yarn` directory in home but this should hopefully prevent it from happening so we can re-enable neuvector scanning.

but it's in the container that matters, not on the Jenkins agent, I don't see how this is relevant

PR #1207, opened 2024-02-22
DTSPO-15800: Angular Version Nagger

reespozzi:
https://tools.hmcts.net/jira/browse/DTSPO-15800

Pipeline nagger for Angular version - checks for angular/core version through package.json files, and compares major version to that listed in deprecation map.

Deprecation map altered in https://github.com/hmcts/cnp-jenkins-config/pull/1034 to include Angular


reespozzi:
Example of a blocked build https://build.hmcts.net/blue/organizations/jenkins/HMCTS_d_to_i%2Fem-media-viewer/detail/PR-943/10/pipeline

Does not block if version is compatible and within current date
Does not affect builds that do not use Angular

danielwilsonkainos (dismissed):

danielwilsonkainos (approved):

PR #1217, opened 2024-03-28
DTSPO-15971 - update tfcmt config

endakelly:
https://tools.hmcts.net/jira/browse/DTSPO-15971

Fix issue with comments being overwritten when running against aat/stg as well as prod.

Comments should now get created per environment and they will be independently updated based on changes in each environment being targeted.

Example PR showing this: https://github.com/hmcts/sds-toffee-recipes-service/pull/434


thomast1906 (approved):
Nice

PR #1221, opened 2024-04-02
Increase warning date for single servers BSP

joshblackmoor:
Notes:
* Increased date for single server warning for BSP so pipeline runs, we are in the process of deleting single servers now as migration is complete in prod.


thomast1906 (changes requested):
You are increasing date for all the repos - is that needed?

joshblackmoor:
> You are increasing date for all the repos - is that needed?

Have split BSP into its own for date

amstevenson (approved):

thomast1906 (approved):

PR #1222, opened 2024-04-03
DTSPO-17287: Switch to using yarn

reespozzi:
Resolves projects with node_modules dir outside of root

### DIFF
--- a/resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
+++ b/resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
@@ -6,7 +6,7 @@ REQUIRED_VERSION=${2}
 
 # Attempt to find angular version in a package.json file and trims version of angular/core to get the major version value 
 CURRENT_VERSION=""
-version=$(yarn info "@angular/core" --json | jq '.children.Version' | tr -d '"' | cut -d '.' -f 1 )
+version=$(yarn info "@angular/core" --json | jq -r '.children.Version' | cut -d '.' -f 1 )
 if [[ -n "$version" ]]; then
     CURRENT_VERSION="$version"
     echo "Current version: $CURRENT_VERSION"
--- a/resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
+++ b/resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
@@ -6,7 +6,7 @@ REQUIRED_VERSION=${2}
 
 # Attempt to find angular version in a package.json file and trims version of angular/core to get the major version value 
 CURRENT_VERSION=""
-for file in $(find . -type f -name "package.json" -not -path "./node_modules/*"); do
+for file in $(find . -type f -name "package.json" -not -path "*/node_modules/*" -not -path "*/dist/*"); do
     version=$(cat "$file" |  grep "$DEPENDENCY" | sed 's/.*"\(.*\)".*/\1/' | tr -d '^' | cut -d '.' -f 1 )
     if [[ -n "$version" ]]; then
         CURRENT_VERSION="$version"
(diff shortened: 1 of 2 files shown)


resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
```diff
@@ -6,7 +6,7 @@ REQUIRED_VERSION=${2}
 
 # Attempt to find angular version in a package.json file and trims version of angular/core to get the major version value 
 CURRENT_VERSION=""
-for file in $(find . -type f -name "package.json" -not -path "./node_modules/*"); do
+for file in $(find . -type f -name "package.json" -not -path "*/node_modules/*" -not -path "*/dist/*"); do
```
timja, line 9: why does it need to check outside of the package.json at the root? is there any projects where the frontend isn't in the root package.json?

timja: and why can't you use the CLI, to get the version instead of all that sed, tr and cut =/

https://stackoverflow.com/a/70254297

reespozzi: em-showcase flagged this in DTSPO-17287

reespozzi: I could get the angular cli version, but struggled to get the angular core version

reespozzi: I also hadn't used yarn info as I couldn't tell if that would be suitable for every single repo, in case some didn't use yarn

reespozzi: I've added it in https://github.com/hmcts/cnp-jenkins-library/pull/1222/commits/5de0de3341b0695d6cd69627bc2a1a7b0bc66e07 -- will re-run against a few apps I know are using angular too

reespozzi: @timja  example of this running too with those suggested yarn changes: https://build.hmcts.net/job/HMCTS_d_to_i/job/em-showcase/job/PR-498/6/consoleFull

timja: > DTSPO-17287

aware but why is it checking anything outside of the root package.json

> I also hadn't used yarn info as I couldn't tell if that would be suitable for every single repo, in case some didn't use yarn

Its a pipeline requirement to use yarn, any Jenkins pipeline repository will be using it.

>  that also returns a few different angular/core versions which are child dependencies too

I'd expect it to be one at the root level but would need testing.

reespozzi: ` pipeline requirement to use yarn` - Wasn't aware of this sorry, I do see now there's a small bit in the readme though

Are you able to re-review please @timja ? Pushed some changes to use yarn now and it is only checking the root level

resources/uk/gov/hmcts/npm/check-deprecated-npm-dependency.sh
```diff
@@ -6,17 +6,13 @@ REQUIRED_VERSION=${2}
 
 # Attempt to find angular version in a package.json file and trims version of angular/core to get the major version value 
 CURRENT_VERSION=""
-for file in $(find . -type f -name "package.json" -not -path "./node_modules/*"); do
-    version=$(cat "$file" |  grep "$DEPENDENCY" | sed 's/.*"\(.*\)".*/\1/' | tr -d '^' | cut -d '.' -f 1 )
-    if [[ -n "$version" ]]; then
-        CURRENT_VERSION="$version"
-        echo "Current version: $CURRENT_VERSION"
-        break
+version=$(yarn info "@angular/core" --json | jq '.children.Version' | tr -d '"' | cut -d '.' -f 1 )
```
timja, line 9:
```suggestion
version=$(yarn info "@angular/core" --json | jq -r '.children.Version')
```

reespozzi: Ah nice thanks -- I still need cut because deprecation map is only looking at the major version https://github.com/hmcts/cnp-jenkins-config/blob/master/deprecation-config.yml#L35

timja (approved):

PR #1225, opened 2024-04-16
Use amd64 zaproxy image from ghcr.io

ieuanb74:
Notes:
* zaproxy also host on gh where they offer amd64 images. This PR updates the image to pull from ghcr.io instead of docker hub
* Fixes https://tools.hmcts.net/jira/browse/DTSPO-17373

### DIFF
--- a/src/uk/gov/hmcts/contino/SecurityScan.groovy
+++ b/src/uk/gov/hmcts/contino/SecurityScan.groovy
@@ -2,7 +2,7 @@ package uk.gov.hmcts.contino
 
 
 class SecurityScan implements Serializable {
-    public static final String OWASP_ZAP_IMAGE = 'owasp/zap2docker-stable:2.13.0'
+    public static final String OWASP_ZAP_IMAGE = 'ghcr.io/zaproxy/zaproxy:20240402-stable@sha256:3280adc730131f1f4460ab226b0f85e3e9ab3301ef5a7030f745ac4dd6b6ff87'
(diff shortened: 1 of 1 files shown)


src/uk/gov/hmcts/contino/SecurityScan.groovy
```diff
@@ -2,7 +2,7 @@ package uk.gov.hmcts.contino
 
 
 class SecurityScan implements Serializable {
-    public static final String OWASP_ZAP_IMAGE = 'owasp/zap2docker-stable:2.13.0'
+    public static final String OWASP_ZAP_IMAGE = 'ghcr.io/zaproxy/zaproxy:20240402-stable@sha256:3280adc730131f1f4460ab226b0f85e3e9ab3301ef5a7030f745ac4dd6b6ff87'
```
endakelly, line 5: Do they still have ARM based images as well? Not a massive issue but AMD64 images may not work on ARM based jenkins agents. That will only affect plum/toffee but might break the pipeline tests

ieuanb74: Yeah so they have arm images too but that was causing security scans to fail as they were running on x64 agents. Could either update the image to use amd64 version or would need teams to specify in their jenkinsfile to use arm agent?

endakelly: It should pick up the right version based on the architecture of the agent without anything else being needed. All the teams are using the x64 agents, it's just our test apps using the ARM agents. I'll approve this for now to get the issue fixed and then can you look into getting the arm one working as well?

endakelly (approved):

PR #1226, opened 2024-04-19
DTSPO-15971 run tf plan on prod by default 

charlesluokb:

DTSPO-15971
- run TF plan against prod by default on PRs
- users can opt out to specify 'not-plan-on-prod' as topic or label 

Notes:
*  could not write unit tests for 'withPipeline` as `gradlew test` is broken at moment
*
*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,9 +13,6 @@ repositories {
 test {
   systemProperty 'groovy.grape.enable', 'false'
   useJUnitPlatform()
-  testLogging {
-    events "passed", "skipped", "failed", "standardOut", "standardError"
-  }
 }
 
 compileGroovy {
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,7 @@ sourceSets {
   }
   test {
     groovy {
-      srcDirs = ['test','test/withPipeline']
+      srcDirs = ['test']
     }
     resources {
       srcDirs = ['testResources']
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,9 @@ repositories {
 test {
   systemProperty 'groovy.grape.enable', 'false'
   useJUnitPlatform()
+  testLogging {
+    events "passed", "skipped", "failed", "standardOut", "standardError"
+  }
 }
 
 compileGroovy {
@@ -61,10 +64,11 @@ sourceSets {
   }
   test {
     groovy {
-      srcDirs = ['test']
+      srcDirs = ['test','test/withPipeline']
     }
     resources {
       srcDirs = ['testResources']
     }
+
   }
 }
--- a/vars/withPipeline.groovy
+++ b/vars/withPipeline.groovy
@@ -126,7 +126,7 @@ def call(type, String product, String component, Closure body) {
             def targetBranch = githubApi.refreshPRCache() // e.g. demo, perftest, ithc, master, or non-standards
             def labelCache = githubApi.refreshLabelCache()
             def topicCache = githubApi.refreshTopicCache()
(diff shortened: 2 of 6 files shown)


vars/withPipeline.groovy
```diff
@@ -107,65 +106,42 @@ def call(type, String product, String component, Closure body) {
...
-              appPipelineConfig: pipelineConfig,
-              pipelineCallbacksRunner: callbacksRunner,
-              pipelineType: pipelineType,
-              subscription: subscription.nonProdName,
-              aksSubscription: aksSubscriptions.aat,
-              environment: environment.nonProdName,
-              product: product,
-              component: component,
-              tfPlanOnly: true
-            )
```
louisehuyton, line 121: so its very essential these lines stay in, we need a tf plan of aat (since its the staging environment) to run every time as a separate step no matter what the topic/label is

vars/withPipeline.groovy
```diff
@@ -107,65 +106,42 @@ def call(type, String product, String component, Closure body) {
...
-
+          onTerraformChangeInPR {
             def githubApi = new GithubAPI(this)
-
+            def prCache = githubApi.refreshPRCache() // e.g. demo, perftest, ithc, etc
```
louisehuyton, line 111: so we need the for loop to go through these values because the user could name their branch anything it doesn't have to be an env name. For example: this_is_a_branch would try perform a tf plan against that value thinking its an environment

vars/withPipeline.groovy
```diff
@@ -14,7 +14,6 @@ import uk.gov.hmcts.contino.PipelineCallbacksConfig
 import uk.gov.hmcts.contino.PipelineCallbacksRunner
 import uk.gov.hmcts.pipeline.AKSSubscriptions
 import uk.gov.hmcts.pipeline.TeamConfig
-import uk.gov.hmcts.pipeline.DeprecationConfig
```
louisehuyton, line 17: this needs to stay - please add back in

build.gradle
louisehuyton, line 1: could you remove the entire file from the pr change please as this is unrelated

vars/withPipeline.groovy
```diff
@@ -107,65 +106,42 @@ def call(type, String product, String component, Closure body) {
...
-              product: product,
-              component: component,
-              tfPlanOnly: true
-            )
-
+          onTerraformChangeInPR {
             def githubApi = new GithubAPI(this)
-
+            def prCache = githubApi.refreshPRCache() // e.g. demo, perftest, ithc, etc
+            def branchName = branch.branchName.toLowerCase() // could be PR-123 or #58
```
louisehuyton, line 112: we use PR-123 instead of branchName because thats how it recognises the format of a PR. I don't think this is necessary.

vars/withPipeline.groovy
louisehuyton: think this may be fine - marking as resolved

build.gradle
charlesluokb: can not remove the entire file but I can revert the change I made in build.gradle. The file's already existed

vars/withPipeline.groovy
charlesluokb: The `DeprecationConfig` function is not referenced in the file hence I removed it. Can I know why it needs to stay

vars/withPipeline.groovy
charlesluokb: done

vars/withPipeline.groovy
charlesluokb: add it back in

endakelly (approved):

PR #1238, opened 2024-05-31
DTSPO-17348: Use new deprecation map location

reespozzi:
As part of https://tools.hmcts.net/jira/browse/DTSPO-17348 we require a central location for nagger details and an update to jenkins functionality for terraform nagger

Previously:

- Hardcoded value for major azurerm provider version
- Hardcoded value for major terraform version 
- slack message sent for each deprecation

Changes:

- This PR updates jenkins-library to use a new central deprecation map location shared by ADO
- Adds functionality for us to dynamically compare terraform versions and terraform provider versions to app configuration (no longer just major versions and no longer just azurerm provider)
- Adds some updates for slack messages for terraform deprecations so it's nicely formatted

All versions and dates are identical as the old mapping for helm / gradle / npm versions
New deprecation map location https://github.com/hmcts/cnp-deprecation-map/pull/3

All naggers for terraform have a date of `"2024-12-30"` for now too, so even if there is outdated repos they will not be broken

Because of the way the Pipeline Warning collector work (only accepting one date) - I have chosen to adopt the soonest possible deprecation date to put in the warning message in the case where more than one deprecation is noticed, so that the statement "Your configuration will stop working by X date" is still technically true. 
All the warnings are collected into a formatted string whilst we go over the deprecation map and then sent in one terraform deprecation slack message as below. This is grouped per env now. The other option was to add a pipeline warning for each deprecation but that got a little bit repetitive in the message, and I think grouping looks nicer

- Example slack message where both terraform version and a provider version are below the required version:
image

- Example slack message where the terraform version will need updating:
image

### DIFF
--- a/vars/warnAboutDeprecatedChartConfig.groovy
+++ b/vars/warnAboutDeprecatedChartConfig.groovy
@@ -30,37 +30,31 @@ def call(Map<String, String> params) {
 
   sh 'chmod +x check-deprecated-charts.sh'
 
-  helmChartDeprecationConfig.each { chart, deprecatedVersions ->
-    deprecatedVersions.each { deprecation ->
-      try {
-        sh "./check-deprecated-charts.sh $product $component $chart $deprecation.version "
-      } catch (ignored) {
-        WarningCollector.addPipelineWarning("deprecated_helmcharts", "Version of $chart helm chart below $deprecation.version is deprecated, please upgrade to latest release https://github.com/hmcts/chart-$chart/releases", LocalDate.parse(deprecation.date_deadline))
-      }
+  helmChartDeprecationConfig.each { chart, deprecation ->
+    try {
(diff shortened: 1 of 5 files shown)


vars/warnAboutDeprecatedChartConfig.groovy
reespozzi, line 1: The changes in this file do not affect functionality, they are because the format of the deprecation map has been slightly altered

danielwilsonkainos (approved):

reespozzi:
Going to try to move [Jenkins nagger](https://www.jenkins.io/doc/pipeline/steps/slack/#slacksend-send-slack-message) to [blocks](https://api.slack.com/reference/block-kit/blocks#blocks) implementation in TD instead of part of this work as there seems to a be a large re-write required

PR #1242, opened 2024-06-17
DTSPO-17456 Run tests for global functions

charlesluokb:
Previously all of the `Jenkinsfile` based tested were being silently skipped.

This happened when groovy was upgraded in a renovate test, in the `groovy-all` artifact there's a testing artifact, in version 2.5 this was switched from `JUnit 4` to `JUnit 5`. There was no engine on the classpath for `JUnit 4` tests and then these were silently being skipped.

These haven't been running for quite awhile so there's been a number of changes were additions haven't been made to the tests or syntax has been used that doesn't work in the pipeline library.

The most common problem was type information breaking the tests, a `ClassCastException` or a `VerifyError` would be thrown even though all the inputs matched what should be passed to the function or constructor.

The way to fix this was to drop the type information 😢 

It might be possible to use a newer groovy and spock version e.g. see https://github.com/jenkinsci/JenkinsPipelineUnit/issues/499#issuecomment-1070816337. But its probably safer to leave it as 2.4 as that is what Jenkins is running on and someone might try and use a newer groovy feature than Jenkins supports


build.gradle
```diff
@@ -13,32 +13,81 @@ repositories {
...
   implementation 'org.jenkins-ci.plugins:job-dsl-core:1.87'
   implementation group: 'org.jenkins-ci.plugins', name: 'scm-api', version: '616.ve67136f6c77d', ext: 'jar'
 
-  testImplementation group: 'org.spockframework', name: 'spock-core', version: '2.3-groovy-2.5'
+  testImplementation 'junit:junit:4.13.2'
+  testImplementation group: 'org.spockframework', name: 'spock-core', version: '1.3-groovy-2.4'
   testImplementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.16.0'
 
-  testImplementation group: 'com.lesfurets', name: 'jenkins-pipeline-unit', version: '1.17'
+  testImplementation 'org.junit.vintage:junit-vintage-engine:5.10.2'
```
charlesluokb, line 87: even downgraded groovy to 2.4.21, we still need to declare the Junit Vintage dependency

timja: we don't you'd configured junit to use junitPlatform at this point

build.gradle
```diff
@@ -12,33 +12,61 @@ repositories {
...
+      ", Failures: ${results.sum { it.failedTestCount }}" +
+      ", Errors: ${results.sum { it.exceptions.size() }}" +
+      ", Skipped: ${results.sum { it.skippedTestCount }}"
+  }
 }
 
 compileGroovy {
   groovyOptions.forkOptions.jvmArgs = [ '-Dgroovy.grape.enable=false' ]
+  // @CNP annotation comes from the Jenkins Runtime instance
+  exclude '**/pluginActive.groovy'
```
timja, line 40: is this needed?

build.gradle
```diff
@@ -12,33 +12,61 @@ repositories {
...
+  }
 
   implementation group: 'com.cloudbees', name: 'groovy-cps', version: '3773.v505e0052522c'
   testImplementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.12.0'
 
-  implementation group: 'org.jenkins-ci.main', name: 'jenkins-core', version: '2.409'
-  implementation group: 'org.jenkins-ci.plugins.workflow', name: 'workflow-step-api', version: '625.vd896b_f445a_f8', ext: 'jar'
-  implementation group: 'org.jenkinsci.plugins', name: 'pipeline-model-definition', version: '2.2169.vee7cd0efc13e', ext: 'jar'
+  implementation (group: 'org.jenkins-ci.main', name: 'jenkins-core', version: '2.460') {
+      exclude group: "org.codehaus.groovy", module: "groovy-all"
```
timja, line 56: hmm we can probably just pull the groovy dependency from Jenkins and remove the dependency on line 48?

PR #1249, opened 2024-07-02
Increase helm timeout - temp help for ASO issues

hannah38:
Linked to issues on Preview with Azure Service Operator.
Azure resources (postgres db/sb topics/sb queues/storage container etc).
Want to increase this timeout so more pass and we have less cycles of not deploy quick enough, be deleted again and retry deploying. Hope is that this leads to more passing first time, and then there is less pressure on ASO/Azure.

We will still continue with the Msft support ticket and ASO team on slack.

From some recent logs from ASO we are seeing a big gap between Reconcile Invoked and the About to send resource to Azure log. And the resources end up being deleted soon after they start deploying because of this timeout.

Notes:
*
*
*


thomast1906 (dismissed):
this enough time for a test?

cpareek (approved):

PR #1256, opened 2024-07-16
DTSPO-18118: Update slackSend to use blocks

reespozzi:
### Overview
Moves slack messaging from static messages to Blocks via attachments (in this case using attachments as color not available directly with blocks -- although Blocks as a whole is recommended, you can still pass these blocks through in attachments, [see these docs](https://api.slack.com/messaging/attachments-to-blocks#direct_equivalents))
```
There is one exception, and that's the [color parameter](https://api.slack.com/reference/messaging/attachments#fields), which currently does not have a block alternative. If you are strongly attached (🎺) to the color bar, use the [blocks parameter](https://api.slack.com/reference/messaging/attachments#fields) within an attachment.

This will let you place newer blocks inside of an attachment
``` 
This^ results in less noticeable change to an end user.

- Unfortunately we may have to lose having the same header for the terraform nagger, as there's a 150 character limit and URLs are not formatted nicely. We can't have [ links in Headers](https://api.slack.com/reference/block-kit/blocks#header) I've kept the header but moved the links to a footer section.

- This change is to keep up to date with more modern features and extensibility if we ever want it in the future.
  ADO pipeline terraform nagger was also recently implemented to send slack messages using blocks.

- This way is the path of least resistance as any other method I though of involves rewriting the WarningCollector (which is used in countless places). This just manipulates what's already in place into a correct format before any `slackSend` function, of which there's only 4

### Changes:
- New class `SlackBlockMessage` so we can manipulate the object to send to slack more easily from everywhere
- Updates tests to handle new functionality

### Functionality:
I've kept it as close to original as possible whilst allowing for new changes. This is a starting point / migration so if we want in the future we can make more changes / add functionality 

Example of new slackSend working for buildFixed and buildFailed whilst retaining color:
image

Example for notifyPipelineDeprecations (any deprecations) -- largely similar to before but link had to go at the bottom to be formatted nicely
image


Jenkinsfile_pipeline_test
```diff
@@ -1,5 +1,5 @@
 #!groovy
-@Library('Infrastructure') _
+@Library('Infrastructure@DTSPO-18118-jenkins-block-messages') _
```
MartyFox, line 2: revert back to master

MartyFox (dismissed):

NickAzureDevops (approved):

PR #1280, opened 2024-11-07
Allow Java21 on yarn builds

reespozzi:
Weird edge case where a nodejs pipeline requires a java based job

On a typical java build, we have this logic during that groovy call: https://github.com/hmcts/cnp-jenkins-library/blob/034812d8daebc9396c43e48ae3d6333b60083a31/src/uk/gov/hmcts/contino/GradleBuilder.groovy#L250-L272
But your build here uses yarn, which uses [YarnBuilder](https://github.com/hmcts/cnp-jenkins-library/blob/034812d8daebc9396c43e48ae3d6333b60083a31/src/uk/gov/hmcts/contino/YarnBuilder.groovy) — my change on the branch added a java -version lookup https://github.com/hmcts/cnp-jenkins-library/compare/master...java-home-for-node-builds — if you look at the console log of that PR 444 i opened, you’ll see this:

```
16:11:53  + echo java_version_lookup
16:11:53  java_version_lookup
16:11:53  [Pipeline] sh
16:11:53  + java -version
16:11:53  openjdk version "17.0.12" 2024-07-16
16:11:53  OpenJDK Runtime Environment (build 17.0.12+7-Ubuntu-1ubuntu220.04)
16:11:53  OpenJDK 64-Bit Server VM (build 17.0.12+7-Ubuntu-1ubuntu220.04, mixed mode, sharing)
```

This change allows users to use java21 in Yarn Builds if they wish to

Probably there's not many other cases using java on node builds, but I still made this conditional so it doesn't break anything once merged

### DIFF
--- a/src/uk/gov/hmcts/contino/YarnBuilder.groovy
+++ b/src/uk/gov/hmcts/contino/YarnBuilder.groovy
@@ -397,11 +397,11 @@ EOF
     super.setupToolVersion()
     // Java required on nodejs pipeline, if data import only available as java job, but project itself is nodejs
     def statusCodeJava21 = steps.sh script: 'grep -F "JavaLanguageVersion.of(21)" build.gradle', returnStatus: true
-    if (statusCodeJava21 == 0) {
-      def javaHomeLocation = steps.sh(script: 'ls -d /usr/lib/jvm/temurin-21-jdk-*', returnStdout: true, label: 'Detect Java location').trim()
-      steps.env.JAVA_HOME = javaHomeLocation
-      steps.env.PATH = "${steps.env.JAVA_HOME}/bin:${steps.env.PATH}"
-    }
+    localSteps.sh "echo setting_java_version"
+
+    def javaHomeLocation = steps.sh(script: 'ls -d /usr/lib/jvm/temurin-21-jdk-*', returnStdout: true, label: 'Detect Java location').trim()
+    steps.env.JAVA_HOME = javaHomeLocation
+    steps.env.PATH = "${steps.env.JAVA_HOME}/bin:${steps.env.PATH}"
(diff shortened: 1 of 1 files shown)


src/uk/gov/hmcts/contino/YarnBuilder.groovy
```diff
@@ -395,6 +395,15 @@ EOF
   @Override
   def setupToolVersion() {
     super.setupToolVersion()
+    // Java required on nodejs pipeline, if data import only available as java job, but project itself is nodejs
+    def statusCodeJava21 = steps.sh script: 'grep -F "JavaLanguageVersion.of(21)" aat/build.gradle', returnStatus: true
```
rapolaskaseliscgi, line 399: It is not necessarily lives in "aat/build.gradle" location, example here https://github.com/hmcts/civil-ccd-definition , build.gradle is at the root of project, not under "aat" subdirectory.

reespozzi: Is that build also using a java job on nodejs?

reespozzi: [2cd0362](https://github.com/hmcts/cnp-jenkins-library/pull/1280/commits/2cd0362c3eeac7aa3b6a08c82d0111cc8ba5b0f8) should be checking any that have 21

rapolaskaseliscgi: Yes, that build in my understanding is similar that is nodejs using java job. After looking in more repos (there aren't many, maybe 5), I'd say that `build.gradle` more likely to be at the root level, and not under aat subdirectory.

PR #1290, opened 2025-01-09
Dtspo 17456 run jenkinsfile based tests being silently skipped

charlesluokb:
Jira Ticket: https://tools.hmcts.net/jira/browse/DTSPO-17456

## The problem
All of Jenkinsfile based unit tests (e.g. withJavaPipelineXxxTests, etc) are being silently skipped.

This happened when groovy was upgraded in a renovate test, in the groovy-all artifact there's a testing artifact, in version 2.5 this was switched from JUnit 4 to JUnit 5. There was no engine on the classpath for JUnit 4 tests and then these were silently being skipped.
These haven't been running for quite awhile so there's been a number of changes were additions haven't been made to the tests or syntax has been used that doesn't work in the pipeline library.

### Variable Type Problem

This problem caused a `ClassCastException` or `VerifyError` being thrown even though all the inputs matched what should be passed to the function or class constructor. 
The way to fix this is to drop the type information and use `def` instead

### Pipeline Steps Evolving

Since `groovy-all` was upgraded to 2.5 there's been a number of changes (step modifications, library syntax, etc) were introduced and haven't been made to the tests, a massive technical debt is accumulated in the areas such as test setup, stubs setup, modulising these shared libraries (i.e. vars/*.groovy)

#### Pipeline Step Stubs

An error likes `Fails with verify[0]: expected 1..1 call(s) to 'setupToolVersion' but was called 0 time(s)` being thrown in many test cases `PipelineExecutesExpectedSteps` or `PipelineExecutesExpectedStepsInExpectedOrderWithSkips` I could not figure it out within the time limit. 

#### null `currentBuild` object injected 

An exception below is thrown in `notifyBuildFixed.groovy` and I also could not figure it out why
``` java.lang.Exception: ERROR: Failed to notify #product-builds due to the following error: groovy.lang.MissingMethodException: No signature of method: java.util.LinkedHashMap.getPreviousBuild() is applicable for argument types: () values: [] ```

### CODE REVIEW

File: Jenkinsfile_pipeline_test [hunk 1]
```diff
@@ -1,5 +1,5 @@
 #!groovy
-@Library('Infrastructure') _
+@Library('Infrastructure@DTSPO-17456-runJenkinsfileBasedTestsBeingSilentlySkipped') _
```
charlesluokb, line 2: will revert to point to `master`

src/uk/gov/hmcts/contino/GradleBuilder.groovy
```diff
@@ -15,13 +15,10 @@ class GradleBuilder extends AbstractBuilder {
   // https://issues.jenkins.io/browse/JENKINS-47355 means a weird super class issue
   def localSteps
 
-  def securitytest
-
   GradleBuilder(steps, product) {
     super(steps)
     this.product = product
     this.localSteps = steps
-    this.securitytest = new SecurityScan(this.steps)
```
charlesluokb, line 24: removed as it is a duplicate call. The call already exists in the super class `AbstractBuilder.groovy` and is implicitly called on the line 19 `super(steps)` f.y.i.  @reespozzi

reespozzi (approved):
Both check pipelines passed, and a test repo too

PR #1300, opened 2025-02-20
extend terraform provider from default

thomast1906:
Using same pattern as per `warnAboutDeprecatedPostgres.groovy`

- Similar was added previous for https://github.com/hmcts/cnp-jenkins-library/blob/master/vars/warnAboutDeprecatedPostgres.groovy

### DIFF
--- a/vars/warnAboutOldTfAzureProvider.groovy
+++ b/vars/warnAboutOldTfAzureProvider.groovy
@@ -15,8 +15,8 @@ def call(String environment, String product) {
       ? LocalDate.of(2024, 7, 2)
       : LocalDate.parse(deprecation.date_deadline)
     
-    if (LocalDate.now().isAfter(deadlineDate)) {
-      error "Dependency ${dependency} has expired on ${deadlineDate}. Please upgrade to version ${deprecation.version} or higher."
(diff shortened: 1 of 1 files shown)


vars/warnAboutOldTfAzureProvider.groovy
```diff
@@ -3,6 +3,18 @@ import uk.gov.hmcts.pipeline.DeprecationConfig
 import java.time.LocalDate
 
 def call(String environment, String product) {
+  String gitUrl = env.GIT_URL
+  LocalDate defaultExpiryDate = LocalDate.parse(deprecation.date_deadline)
+
+  // Repository specific expiry dates
+  switch (gitUrl.toLowerCase()) {
+    case "https://github.com/hmcts/am-org-role-mapping-service.git":
```
thomast1906, line 11: repo to be changed for associated repo we want to extend

vars/warnAboutOldTfAzureProvider.groovy
```diff
@@ -3,8 +3,17 @@ import uk.gov.hmcts.pipeline.DeprecationConfig
 import java.time.LocalDate
 
 def call(String environment, String product) {
-
+  String gitUrl = env.GIT_URL
```
bancey, line 6: is this `gitUrl` var used?

PR #1326, opened 2025-07-29
Fix all tests that were ignored

DanielKChristou:
- Switches to JsonSlurperClassic to avoid type issues when parsing in Jenkins.

- Add null and method checks to avoid failures when Jenkins objects don’t behave as expected(usually when running code as part unit tests).

- Makes the previous build check safer by only calling getPreviousBuild() if it exists.

- Updates tests to reflect the safer parsing and null handling.

- Removes reliance on Groovy’s dynamic typing to avoid issues with newer versions.

### DIFF
--- a/test/withNightlyPipeline/onMaster/withAngularNightlyPipelineOnMasterTests.groovy
+++ b/test/withNightlyPipeline/onMaster/withAngularNightlyPipelineOnMasterTests.groovy
@@ -1,7 +1,6 @@
 package withNightlyPipeline.onMaster
 
 import org.junit.Test
-import uk.gov.hmcts.contino.AngularBuilder
 import withPipeline.BaseCnpPipelineTest
 
 class withAngularNightlyPipelineOnMasterTests extends BaseCnpPipelineTest {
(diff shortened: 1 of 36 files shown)


vars/notifyBuildFixed.groovy
```diff
@@ -35,7 +35,9 @@ def call(Map args = [:]) {
   }
 
   try {
-    if (currentBuild.getPreviousBuild()?.getResult() == 'FAILURE') {
+    // Check if currentBuild has the getPreviousBuild method (may not exist in test environments)
+    def hasPreviousBuild = currentBuild?.hasProperty('getPreviousBuild') || currentBuild?.metaClass?.respondsTo(currentBuild, 'getPreviousBuild')
```
DanielKChristou, line 39: Added checks for null and method existence to prevent runtime errors when Jenkins objects don't have expected properties or methods like getPreviousBuild.

test/withNightlyPipeline/onMaster/withNodeJsNightlyPipelineOnMasterTests.groovy
```diff
@@ -16,20 +14,20 @@ class withNodeJsNightlyPipelineOnMasterTests extends BaseCnpPipelineTest {
 
   @Test
   void NightlyPipelineExecutesExpectedStepsInExpectedOrder() {
-    def mockBuilder = new MockFor(YarnBuilder)
-    mockBuilder.demand.with {
+    def stubBuilder = new StubFor(YarnBuilder)
+    stubBuilder.demand.with {
```
DanielKChristou, line 18: Changed from MockFor to StubFor to fix Groovy 2.5.x compatibility issues. The strict mocking of MockFor  was causing the StackOverflow.

test/withNightlyPipeline/onMaster/withAngularNightlyPipelineOnMasterTests.groovy
```diff
@@ -16,20 +14,29 @@ class withAngularNightlyPipelineOnMasterTests extends BaseCnpPipelineTest {
 
   @Test
   void NightlyPipelineExecutesExpectedStepsInExpectedOrder() {
-    def mockBuilder = new MockFor(AngularBuilder)
-    mockBuilder.demand.with {
+    // Register the saucePublisher method for the test environment
+    helper.registerAllowedMethod("saucePublisher", [], {})
```
DanielKChristou, line 18: Methods were added to make sure all called methods are registered.

charlesluokb (approved):

PR #1331, opened 2025-09-08
Remove db check stage to unblock teams

reespozzi:
MS issue dropped db availability to 0 since friday and is still unresolved. They have agreed this is an issue with backends on their side.


reespozzi:
Showed a green pipeline run of the important stage of this; 
image


merging to unblock teams

PR #1428, opened 2026-04-07
OnlyDeploy option

nilay913:
Notes:
* Adds a config option pulled from GitHub labels to only deploy skipping all tests, dependencies and coverage checks, ultimately failing the build after deployment (so it can't be merged without running the full pipeline) for ease of deployment for development/testing for PoCs/spikes


niallk (dismissed):

thomast1906 (changes requested):
I dont think we will be approving this. 

Bypassing security checks even for PoC I would not recommend, happy to discuss further on Slack.

PR #1520, opened 2026-08-14
Allow smoke tests to skip team secrets

jasonpaige:
## What changed

- add a backwards-compatible `disableSmokeTestSecrets()` pipeline option
- run the smoke-test stage outside `withTeamSecrets` only when that option is enabled
- keep deployment, functional, performance, and other test secret loading unchanged
- document the option and cover both the default and opt-out paths with regression tests

## Why

Some smoke tests only use `TEST_URL` and should not receive the application secrets configured by `loadVaultSecrets`. This option scopes those secrets away from the smoke test while preserving existing behaviour by default.

On current `master`, smoke and subsequent test stages share one team-secret binding. If a later functional or performance test needs those secrets, they are still fetched once after smoke; this change therefore improves least-privilege access and stage attribution but does not guarantee a whole-pipeline timing reduction. It also does not bypass the subscription login or secret loading required by deployment and other stages.

## Prerequisite

- #1522 must merge first because downstream pipeline validation reads the permitted library branches from `master`

## Example

```groovy
withPipeline(type, product, component) {
  loadVaultSecrets(secrets)
  disableSmokeTestSecrets()
}
```

## Validation

- `./gradlew check --rerun-tasks` — 336 tests, 0 failures
- regression coverage verifies that smoke and functional stages exist before checking their secret-loading order
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,17 @@ called after each deployment to each environment.
 
 The smoke tests are to be non-destructive (i.e. have no data impact, such as not creating accounts) and a subset of component level functional tests.
 
+If a smoke test only uses `TEST_URL` and does not require the secrets configured by `loadVaultSecrets`, it can avoid loading them for that stage:
+
+```groovy
+withPipeline(type, product, component) {
+  loadVaultSecrets(secrets)
+  disableSmokeTestSecrets()
+}
+```
+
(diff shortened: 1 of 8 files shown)


gunnertwin (approved):
Reviewed, no actionable findings.

PR #1526, opened 2026-08-19
Backport optional smoke-test secrets for 2.4.10

jasonpaige:
## What changed

Backports the optional smoke-test secret handling from #1520 onto the branch used for the 2.4.9 release line. Pipelines retain the existing behaviour by default and can call `disableSmokeTestSecrets()` to run the smoke stage without loading team Key Vault secrets.

## Why

Testing in et-ccd-callbacks showed that avoiding unnecessary smoke-stage secret retrieval reduced the stage by nearly one minute. Keeping this change on the 2.4.9 line allows PlatOps to review it for a possible 2.4.10 release without pulling in unrelated master changes.

## Compatibility

The default remains `smokeTestSecrets = true`, so existing consumers are unaffected. The opt-out path retains the environment-specific executor but bypasses team-secret retrieval for the smoke stage only. Functional and other test stages continue to load their secrets normally.

## Validation

- Focused configuration and preview pipeline tests: 34 passed
- Full library test suite: 400 passed
- `git diff --check` passed

Original change: #1520

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,17 @@ called after each deployment to each environment.
 
 The smoke tests are to be non-destructive (i.e. have no data impact, such as not creating accounts) and a subset of component level functional tests.
 
+If a smoke test only uses `TEST_URL` and does not require the secrets configured by `loadVaultSecrets`, it can avoid loading them for that stage:
+
+```groovy
+withPipeline(type, product, component) {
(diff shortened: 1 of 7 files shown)


gunnertwin (approved):
No actional findings, looks good

PR #1530, opened 2026-09-04
fix(fortify): use real OAuth credentials for library FoD scan runner

niallk:
Newly onboarded repos without a repo-level fortifyScan task fall back to the library's FoD scan runner, which authenticates via OAuth client_credentials. It was using FORTIFY_USER_NAME/FORTIFY_PASSWORD (SSC-style username/PAT from the per-product Key Vault) as the OAuth client_id/client_secret, which Fortify correctly rejects with 401.

- fortifyOnDemandScan: prefer FORTIFY_OAUTH_CLIENT_ID/CLIENT_SECRET, falling back to FORTIFY_USER_NAME/PASSWORD only if unset
- fortifyScan / sectionBuildAndTest: bind withFortifyOAuthSecrets around the scan step itself, not just the post-scan vulnerability report, so the real OAuth credential is actually available when needed

Notes:
*
*
*

### DIFF
--- a/vars/fortifyOnDemandScan.groovy
+++ b/vars/fortifyOnDemandScan.groovy
@@ -73,10 +73,15 @@ EOF
 }
 trap cleanup EXIT
 
-client_id="\${FORTIFY_USER_NAME:-}"
-client_secret="\${FORTIFY_PASSWORD:-}"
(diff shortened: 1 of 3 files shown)


endakelly:
Added @gunnertwin to this for a review as he worked on implementing this iirc

gunnertwin (approved):
`withFortifyOAuthSecrets` looks scoped too widely but otherwise it looks fine to me

PR #1550, opened 2026-09-18
DTSPO-35143: Add nagger for deprecated terraform module branches

manohar-hmcts:
## What

Adds a config-driven nagger that warns teams still pinning terraform modules to migration branches now that those branches are moving back to `master`.

- New `vars/warnAboutModuleBranches.groovy`, modelled on `warnAboutOldLibraryVersion` / `warnAboutDeprecatedPostgres`.
- New resource script `check-module-branches.sh` that greps `*.tf` for `<module>?ref=<deprecated-branch>` pins (handles `git@`/`git::https` forms, `.git` suffix and `//subdir` paths, with a boundary check so similarly-prefixed branch names don't false-match).
- Wired into `spinInfra` alongside the existing terraform naggers, passing `builtFrom` so per-repo `exceptions` overrides in the deprecation map work.

## How it's driven

Deprecated branches, target ref and deadline come from a new `terraform_modules` section in [cnp-deprecation-map/nagger-versions.yaml](https://github.com/hmcts/cnp-deprecation-map/blob/master/nagger-versions.yaml):

```yaml
terraform_modules:
  terraform-module-postgresql-flexible:
    pattern: "DTSPO-30107-additional-postgres-admins"
    version: "master"
    date_deadline: "2026-10-30"
```

The check is a **no-op until that section exists**, so this PR is safe to merge first; a companion cnp-deprecation-map PR activates it. Before the deadline it adds a Slack deprecation warning via `WarningCollector`; after the deadline the pipeline fails (same behaviour as the other naggers).

## Context

~90 references across ~80 hmcts repos currently pin `terraform-module-postgresql-flexible?ref=DTSPO-30107-additional-postgres-admins` (a Jenkins agents migration branch). Note that branch is still 9 commits ahead of the module's `master` — it needs merging back before teams can repoint, so the deadline in the map should be set accordingly.

## Testing

Resource script exercised locally against fixture `.tf` files covering: SSH pin to deprecated branch (fails), `ref=master` (passes), `git::https` + `.git` form (fails), `//subdir` form (fails), similarly-prefixed other branch (passes).

https://tools.hmcts.net/jira/browse/DTSPO-35143


manohar-hmcts:
Pushed a fix after a self-review found a false positive.

`spinInfra` runs this nagger **after** `terraform init`, so `.terraform/modules/` contains the source of every downloaded module. The original grep scanned those too, so a repo correctly pinned to `master` would still be nagged because of a transitive pin inside somebody else's module — something the team cannot fix from their own repo, and which would have failed a compliant build once the deadline passes. Reproduced locally, now fixed by excluding `.terraform` (and switching `-R` to `-r` so symlinked module dirs aren't followed).

Also wrapped the loop in try/finally so the temporary script is still cleaned up when `WarningCollector` throws past the deadline.

Fixture coverage now: SSH pin (fails), `ref=master` (passes), `git::https` + `.git` (fails), `//subdir` (fails), similar-prefix branch name (passes), pin only inside `.terraform` (passes), pin in both root and `.terraform` (fails, and lists only the root file).

Worth noting for reviewers: the existing `check-deprecated-postgres.sh` has the same `.terraform` blind spot — out of scope here, but probably worth a follow-up.

chris47dev (approved):

PR #1549, opened 2026-09-18
Feature/release drafter changelog from agents

gunnertwin:
Resolves https://hmcts.atlassian.net/browse/DTSPO-34653

Notes:
* automate releases and changelog updates

PR #1552, opened 2026-09-18
DTSPO-34661 - update cnp-jenkins-library README after recent library changes

ryanmcalary11:
**Resolves -** https://hmcts.atlassian.net/browse/DTSPO-34661

### Description

Documents the changes made across the CICD Identity & Access Segmentation epic in the `cnp-jenkins-library` README:

- 2.x is now the default library version
- Environment specific agents, and what teams need to change in their Jenkinsfiles
- The new Renovate automated updates
- The `release-drafter` and changelog process

**NOTE** - will only merge after this PR gets merged - https://github.com/hmcts/cnp-jenkins-library/pull/1549


README.md
```diff
@@ -30,7 +36,104 @@ To use this pipeline in your repo, you must import it in a Jenkinsfile
   @Library('Infrastructure')
 ```
 
-### Opinionated app pipeline
+## Library versioning and 2.x
+
+2.x is the default version of this library. `@Library('Infrastructure')` with no version pin resolves to whatever branch your Jenkins instance is configured to track — normally `master` — which is now 2.x, so most pipelines pick it up without any change to their Jenkinsfile.
```
endakelly: We should be clear here that teams should use a version in their Jenkinsfile. The major version will likely change so let's not refer to `2.x`. We should explicitly say that using `Infrastructure` with no version will default to the master branch and risk the pipeline breaking as new features are added

README.md
```diff
@@ -30,7 +36,104 @@ To use this pipeline in your repo, you must import it in a Jenkinsfile
...
+
+The headline change in 2.x is **environment-specific agents and managed identities**. This replaces the previous model, where a single shared identity was implicitly available to any pipeline stage on any agent, regardless of which environment it was targeting or what kind of work it was doing.
+
+### Environment specific agents
+
+Azure work is now routed to agents scoped to the relevant environment, which authenticate using the managed identity for that environment — rather than inheriting broad access simply from where a pipeline stage happens to run.
+
+In practice:
+
+- The opinionated pipelines place their own stages on the correct agent automatically. `withPipeline` and `withInfraPipeline` start on the agent for the branch's primary target environment (`preview` for a PR build, `aat` for `master`, or the branch's own environment for `demo`/`perftest`/`ithc`), and deployment/infrastructure stages hop onto the agent for whichever environment they're deploying to. `withNightlyPipeline` runs entirely on the AAT agent, since that's the only environment it targets.
```
endakelly: Not strictly true for nightly as some do use other environments. It's mostly aat

README.md
```diff
@@ -30,7 +36,104 @@ To use this pipeline in your repo, you must import it in a Jenkinsfile
...
+`withSubscription` now takes the product and the target environment, so the library can schedule the block onto the right agent and authenticate with the right identity:
+
+```groovy
+withSubscription(subscription, product, params.ENVIRONMENT) {
+  // az / terraform commands run here, on an agent scoped to params.ENVIRONMENT,
+  // authenticated as that environment's managed identity
+}
+```
+
+If the environment (or product) can't be resolved, `withSubscription` falls back to the shared Jenkins identity rather than the target environment's — typically on an `ubuntu-ptl` agent, which is what you'll see in the build log. That's the legacy path — pass both explicitly so your pipeline gets environment-scoped scheduling and identity.
```
endakelly: This functionality will be removed in future and cause failures

README.md
ryanmcalary11: Yep makes sense - I'll make sure to update to explicitly say they should be using a specific version and if ran directly against master it will likely cause issues 🙌

README.md
ryanmcalary11: No worries will make sure to note that in the readme 👌 Maybe would it be better to remove that section on the `withSubscription` entirely or wait until it's fully removed?

endakelly: Remove it entirely I think. We should encourage teams to provide the values to the function

manohar-hmcts (dismissed):

manohar-hmcts (approved):

